PR #3, opened 2024-10-01
Prepare 0.9.0 beta version, add dependecy managers

vahancouver:
[SDKS-3361](https://bugster.forgerock.org/jira/browse/SDKS-3361) [iOS] Publish DaVinci SDK to dependency managers

1. Add Package.swift 
2. Add podspec files for all libraries
3. Add PrivacyInfo.xcprivacy to all libraries
4. Add CHANGELOG.md
5. Add THIRD-PARTY-NOTICE.txt
6. Remove PingOrhestrate dependency from PingDavinci as it already depends on PingOidc (which depends on PingOrchestrate)

### DIFF
--- a/PingDavinci.podspec
+++ b/PingDavinci.podspec
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'PingDavinci'
-  s.version          = '0.9.0-beta1'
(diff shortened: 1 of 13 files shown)


spetrov (approved):
Looks good to me!

PR #8, opened 2024-11-13
Sdks 3560 rename modules

vahancouver:
[SDKS-3560](https://pingidentity.atlassian.net/browse/SDKS-3560)

[iOS] Remove "Ping" prefix from modules and classes in the new iOS repo

Removed all `Ping` prefixes form modules and classes, updated folders, podfiles, package.swift and readme files...

Didn't change: PingExample app name, PingTestHost name, Ping workspace name

**I am not 100% sure this is a good approach as there might be conflicts with other module names. Doing more investigation**


spetrov (approved):
Looks good to me! 👍🏻

rodrigoareis (approved):
Changes looks good to me

PR #14, opened 2024-12-02
Fix the issue with HTTP request timeout

vahancouver:
[SDKS-3613](https://pingidentity.atlassian.net/browse/SDKS-3613)
[iOS][Ping SDK] HTTP request timeout config does not work


rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
Looks good to me!

PR #15, opened 2024-12-04
Change requestTimeout implementation

vahancouver:
[SDKS-3613](https://pingidentity.atlassian.net/browse/SDKS-3613)
[iOS][Ping SDK] HTTP request timeout config does not work

This PR changes the request timeout  implementation that doesn't recreate the httpclient session (which had side efefcts)


spetrov (approved):
LGTM! 👍🏻

PR #16, opened 2024-12-04
Sdks 3615 handle 302

vahancouver:
[SDKS-3615](https://pingidentity.atlassian.net/browse/SDKS-3615)
[iOS] Throws Unknown error instead of APIException when server return 302 with empty body

I also fixed the 2 following issues in that branch:
- The subprojects/libraries wouldn’t compile automatically
- Token and user info views would hang the app sometimes.

### DIFF
--- a/SampleApps/PingExample/PingExample/AccessToken.swift
+++ b/SampleApps/PingExample/PingExample/AccessToken.swift
@@ -16,27 +16,26 @@ struct AccessTokenView: View {
     
     var body: some View {
         VStack {
-            
-            TextEditor(text: $accessToken.accessToken)
-                            .foregroundStyle(.secondary)
-                            .padding(.horizontal)
-                            .navigationTitle("AccessToken")
+            ScrollView {
+                Text($accessToken.accessToken.wrappedValue)
+                  .foregroundStyle(.secondary)
+                  .padding(.horizontal)
+                  .navigationTitle("AccessToken")
+            }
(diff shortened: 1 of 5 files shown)


SampleApps/PingExample/PingExample/AccessToken.swift
```diff
@@ -16,27 +16,26 @@ struct AccessTokenView: View {
     
     var body: some View {
         VStack {
-            
-            TextEditor(text: $accessToken.accessToken)
-                            .foregroundStyle(.secondary)
-                            .padding(.horizontal)
-                            .navigationTitle("AccessToken")
+            ScrollView {
```
rodrigoareis, line 19: I believe we decided to rename the folder of sample apps to something like `testApps` or `e2e`, right? Should we do it now?

rodrigoareis (approved):
Changes looks good to me

jeyanthanperiyasamy (approved):

SampleApps/PingExample/PingExample/AccessToken.swift
vahancouver: @rodrigoareis we have decided for the FR repo and there is a ticket in this sprint for that (to name it e2e). I'm not aware of a similar decision for the Pings SDKs.

spetrov (approved):
LGTM! 👍🏻

PR #31, opened 2025-03-04
Sdks 3728 swift 6 support

vahancouver:
[SDKS-3728](https://pingidentity.atlassian.net/browse/SDKS-3728) [iOS] Revisit Swift 6 support (part 1/2)

Update the following modules to support Swift 6 and complete strict concurrency mode
- Logger
- Storage
- Orchestrate

### DIFF
--- a/Orchestrate/Orchestrate/CookieModule.swift
+++ b/Orchestrate/Orchestrate/CookieModule.swift
@@ -2,7 +2,7 @@
 //  CookieModule.swift
 //  PingOrchestrate
 //
-//  Copyright (c) 2024 Ping Identity. All rights reserved.
+//  Copyright (c) 2024 - 2025 Ping Identity. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -29,7 +29,7 @@ public class CookieModule {
         setup.start { context, request in
             let cookies = try? await setup.config.cookieStorage.get()
             if let url = request.urlRequest.url, let cookies = cookies {
-                CookieModule.inject(url: url,
+                await CookieModule.inject(url: url,
                                     cookies: cookies,
                                     inMemoryStorage: setup.config.inMemoryStorage,
                                     request: request)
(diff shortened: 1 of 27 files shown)


Orchestrate/Orchestrate/CookieModule.swift
```diff
@@ -175,19 +182,19 @@ extension Workflow {
 }
 
 /// A storage class for managing in-memory cookies.
-public final class InMemoryCookieStorage: HTTPCookieStorage {
+public actor InMemoryCookieStorage {
```
george-bafaloukas-forgerock, line 185: If we intent to keep this as final I suggest we add `final` here as well. According to Swift forums while actors do not implement inheritance, this might change in the future. Xcode will probably not have an issue with that.
https://forums.swift.org/t/why-can-you-constrain-to-final-classes-and-actors/65256/3

Orchestrate/Orchestrate/HttpClient.swift
```diff
@@ -13,10 +13,10 @@ import Foundation
 import PingLogger
 
 /// `HttpClient` is responsible for handling HTTP requests and logging the details of those requests and responses.
-public class HttpClient {
-    var session: URLSession
+public final class HttpClient: Sendable {
```
george-bafaloukas-forgerock, line 16: Do we need to restrict this to final? Could we benefit from making this an open class for further customization?

george-bafaloukas-forgerock (commented):
Changes look good to me. We need to figure out the tests on the CI to make sure nothing else is broken though. Few questions/comment left for discussion

Orchestrate/Orchestrate/CookieModule.swift
vahancouver: Added the `final` just in case

Orchestrate/Orchestrate/HttpClient.swift
vahancouver: yeah that sounds good. made it open with `@unchecked Sendable`

vahancouver:
> Changes look good to me. We need to figure out the tests on the CI to make sure nothing else is broken though. Few questions/comment left for discussion

@george-bafaloukas-forgerock Addressed the comments and updated the github workflow to use newer xcode and simulator that supports Swift 6. All tests are passing both locally and on github

spetrov (approved):
Looks good to me!

rodrigoareis (approved):
Changes looks good to me

PR #46, opened 2025-04-23
SDKS-3877_Phone_Number_Collector

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-3877) "Support for Phone_Number_Collector"

# Description

# Checklist:

### DIFF
--- a/Davinci/Davinci/collector/PhoneNumberCollector.swift
+++ b/Davinci/Davinci/collector/PhoneNumberCollector.swift
@@ -16,9 +16,9 @@ open class PhoneNumberCollector: FieldCollector<[String: Any]>, @unchecked Senda
     /// validate phone number
     public private(set) var validatePhoneNumber: Bool
     /// country code
-    public var countryCode: String?
+    public var countryCode: String = ""
     /// phone number
-    public var phoneNumber: String?
+    public var phoneNumber: String = ""
 
     /// Initializes a new instance of `PhoneNumberCollector` with the given JSON input.
     public required init(with json: [String : Any]) {
@@ -37,7 +37,7 @@ open class PhoneNumberCollector: FieldCollector<[String: Any]>, @unchecked Senda
     
     /// Returns the selected device type.
     override open func payload() -> [String: Any]? {
-        if countryCode == nil || phoneNumber == nil {
+        if countryCode.isEmpty || phoneNumber.isEmpty{
             return nil
         }
         var payload: [String: Any] = [:]
--- a/Davinci/Davinci/collector/PhoneNumberCollector.swift
+++ b/Davinci/Davinci/collector/PhoneNumberCollector.swift
@@ -9,24 +9,22 @@
 //
 
 /// A collector for phone number.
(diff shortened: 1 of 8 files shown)


Davinci/Davinci/collector/PhoneNumberCollector.swift
```diff
@@ -9,24 +9,30 @@
...
-open class PhoneNumberCollector: FieldCollector<[String: Any]>, Submittable, @unchecked Sendable {
+open class PhoneNumberCollector: FieldCollector<[String: Any]>, @unchecked Sendable {
     
     /// default country code
     public private(set) var defaultCountryCode: String
+    /// validate phone number
+    public private(set) var validatePhoneNumber: Bool
     /// country code
-    public  var countryCode: String?
+    public var countryCode: String?
```
witrisna, line 19: Open for discussion, for Android, I use empty String "" instead of null to represent empty. Other Collector uses empty string as well.
Using empty string, we don't have to do null check everytime we use the attribute.

rodrigoareis: Let's align both platforms, whatever is the preference

george-bafaloukas-forgerock: memory wise, leaving it optional is better, especially when this is a public variable...but the hill I am willing to die on as this has very very low impact and won't change much in this implementation if we do 
`public var countryCode: String = ""`
We already do the same in the SingleValueCollector
`public var value: String = ""`
I will alter this to match Android and the other one. @vahancouver thoughts?

vahancouver: yeah let's keep it consistent, doesn't make a big difference anyway

Davinci/Davinci/collector/PhoneNumberCollector.swift
```diff
@@ -9,29 +9,35 @@
...
     public required init(with json: [String : Any]) {
         self.defaultCountryCode = json[Constants.defaultCountryCode] as? String ?? ""
+        self.validatePhoneNumber = json[Constants.validatePhoneNumber] as? Bool ?? false
         super.init(with: json)
     }
     
-    /// Return event type
-    func eventType() -> String {
-        return Constants.submit
+    /// Initializes the `PhoneNumberCollector` with the given phone number  .
```
vahancouver, line 30: two extra spaces after number

vahancouver (approved):
LGTM. Minor comment

witrisna (approved):

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
LGTM! 👍🏻

Davinci/Davinci/collector/PhoneNumberCollector.swift
george-bafaloukas-forgerock: Done

PR #52, opened 2025-05-16
Sdks 3958 split social collectors separate fb 16 updated names

vahancouver:
[SDKS-3958](https://pingidentity.atlassian.net/browse/SDKS-3958)

iOS-Split the Social Login module to Native and Redirect Components

This PR includes

- All the previous implementation from George
- Downgrading facebook SDK to 16.3.1
- Changing External IdP libraries names to avoid hyphen/underscore
- Updating all ExternalIdP modules Readme files
- Updating Package.swift for the new separate IdP modules
- Adding cocoapods for new separate IdP modules


rodrigoareis (approved):
Changes looks good to me

ExternalIdPFacebook/ExternalIdPFacebook/Handlers/FacebookRequestHandler.swift
```diff
@@ -32,21 +34,27 @@ class FacebookRequestHandler: IdpRequestHandler {
...
         }
     }
     
     /// Initializes a new instance of `FacebookRequestHandler`.
     /// - Parameter httpClient: The `HttpClient` to use for requests
+    @objc(initWithHttpClient:)
     init(httpClient: HttpClient) {
         DispatchQueue.main.async {
             /// Initialize Facebook SDK
+            Settings.shared.isAdvertiserIDCollectionEnabled = true
```
george-bafaloukas-forgerock, line 55: I am not sure we need those for the 16 version of the SDK. But they don't cause any errors let's leave them as they are needed for the newer versions

george-bafaloukas-forgerock:
We need to fix the test cases. Will try to do this today/tomorrow

PR #56, opened 2025-06-13
SDKS-3727_remove hardcoded client ids

george-bafaloukas-forgerock:
Adding DavinciBaseTesta and config json along with changes in the tests to remove hardcoded values

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-3727) "Jira ticket title..."

# Description

# Checklist:

### DIFF
--- a/Davinci/DavinciTests/Configuration/Config.swift
+++ b/Davinci/DavinciTests/Configuration/Config.swift
@@ -1,0 +1,124 @@
+//
+//  Config.swift
+//  Davinci
+//
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+
+import Foundation
+
+@testable import PingDavinci
+
+public enum ConfigError: Error {
+    case emptyConfiguration
+    case invalidConfiguration(String)
+}
+
+class Config: NSObject {
+    var username: String
+    var userFname: String
+    var userLname: String
+    var password: String
+    var newPassword: String
+    var verificationCode: String
+    
+    var clientId: String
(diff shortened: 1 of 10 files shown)


Davinci/DavinciTests/Configuration/Config.swift
```diff
@@ -0,0 +1,124 @@
...
+    var acrValues: String
+    var configPlistFileName: String?
+    
+    var configJSON: [String: Any]?
+    
+    override init() {
+        username = "e2euser@example.com"
+        userFname = "E2E"
+        userLname = "iOS"
+        password = "Demo1234#1"
```
spetrov, line 42: We should not set any sensitive data here...

spetrov (approved):
Looks good to me! 👍🏻 
I can take over from here and update the GitHub actions workflows to initialize the test configuration from the secret store and use it in tests.

vahancouver (approved):
Changes look good. 
Only question is in some cases we replaces the localhost links and in other cases live server links. but we have only one config file. Am I missing something?

george-bafaloukas-forgerock:
> Changes look good. Only question is in some cases we replaces the localhost links and in other cases live server links. but we have only one config file. Am I missing something?

We have the integration tests that use a live env. The rest and in some cases mixed there are mock tests or pointing to test values that don't need to connect. The "sensitive" ones obviously are the live envs

george-bafaloukas-forgerock:
@vahancouver also you can override with a config file or with manually overriding in code like:
self.config.clientId = "test"
        self.config.scopes = ["openid", "email", "address"]
        self.config.redirectUri = "http://localhost:8080"
        self.config.discoveryEndpoint = "http://localhost/.well-known/openid-configuration"
        
        I don't think we need to pollute the config file with test values

PR #60, opened 2025-07-03
SDKS-4189_Cookie_module_updates

george-bafaloukas-forgerock:
Removing logic that deletes the cookies from the in memory storage on every request.

Changing injecting logic to conform to cookie header ";" separated cookies

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4189) "Jira ticket title..."

# Description

# Checklist:


PingTestHost/PingTestHost.xcodeproj/xcshareddata/xcschemes/PingTestHost.xcscheme
```diff
@@ -165,6 +165,66 @@
                ReferencedContainer = "container:../External-idp-Google/ExternalIdPGoogle.xcodeproj">
             </BuildableReference>
          </TestableReference>
+         <TestableReference
```
rodrigoareis, line 168: Do we want to commit this change?

rodrigoareis (approved):
Changes looks good to me. Let's just make sure we want to commit the changes to PingTestHost project files

PingTestHost/PingTestHost.xcodeproj/xcshareddata/xcschemes/PingTestHost.xcscheme
vahancouver: @george-bafaloukas-forgerock there are references to journey module ...

witrisna (changes requested):

spetrov (commented):
LGTM! But I agree with Andy - the `inMemoryStorage` parameter of the `inject` function is no longer used, so we should remove it...

vahancouver (changes requested):
Please see my comment about only deleting one line (_await inMemoryStorage?.deleteCookies(url: url)_) in inject method and preserving the rest of the code.

I also noticed that in validateURL method we are not validating if the url is Secure. I think we should change the method to this
```
func validateURL(_ url: URL) -> Bool {
        return self.validateDomain(url: url) && self.validatePath(url: url) && self.validateIsSecure(url)
    }
```

george-bafaloukas-forgerock:
> Please see my comment about only deleting one line (_await inMemoryStorage?.deleteCookies(url: url)_) in inject method and preserving the rest of the code.
> 
> I also noticed that in validateURL method we are not validating if the url is Secure. I think we should change the method to this
> 
> ```
> func validateURL(_ url: URL) -> Bool {
>         return self.validateDomain(url: url) && self.validatePath(url: url) && self.validateIsSecure(url)
>     }
> ```

I've changed that, but in this case we are basically only supporting `https` urls. Should be mostly OK but this will affect on-prem dev environments potentially. If we do that, we should add a line on the docs that the SDK only supports https URLs

witrisna (approved):

vahancouver (approved):
LGTM

PR #65, opened 2025-07-22
SDKS-4122 Implement Default Device ID Provider for iOS

george-bafaloukas-forgerock:
Adding DeviceIdentifier implementation interfaces and default implementation

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4122) "Implement Default Device ID Provider for iOS"

# Description

# Checklist:


Device/Device/Device.xcodeproj/project.pbxproj
vahancouver, line 1: There is extra Device folder in the hierarchy

vahancouver (changes requested):
Left some comments

Device/Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,232 @@
...
+    public var id: String {
+        get async throws {
+            return try await getIdentifier()
+        }
+    }
+    
+    /// Initializes DeviceIdentifier
+    ///
+    /// - Parameter keychainService: Designated KeychainService to persist, and manage generated Key Pair, and Identifier
+    public init(logger: Logger? = nil) {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: LOL I had that and removed it on a refactoring I did :D

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+
+
+extension Data {
+    /// Converts Data to a hexadecimal string representation.
+    func toHexString() -> String {
+        return map { String(format: "%02hhx", $0) }.joined()
+    }
+}
+
+enum DeviceIndentifierConstants {
```
vahancouver, line 245: enum name misspelled

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        }
+        let hexString = Data(bytes: digest, count: digest.count).toHexString()
+        return hexString
+    }
+    
+    /// Hashes given Data using SHA256 and converts it to a hexadecimal string
+    /// - Parameter data: Data to be hashed
+    /// - Returns: Hashed String of given Data in hexadecimal format
+    /// This is the preferred method to use for hashing device identifiers.
+    func hashAndBase64Data(_ data: Data) -> String {
```
vahancouver, line 143: Same here. The name is misleading as we don't base64 encode here

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+    
+    
+    /// Hashes given Data using SHA1
+    ///
+    /// - Parameter data: Data to be hashed
+    /// - Returns: Hashed String of given Data
+    /// Legacy method, use `hashAndBase64Data(_:)` instead
+    func hashSHA1AndBase64Data(_ data: Data) -> String {
```
vahancouver, line 130: The name is misleading as we don't base64 encode here

vahancouver (changes requested):
Left some comments to address

Device/DeviceTests/DeviceTests.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+
+import XCTest
+@testable import Device
+
+final class DeviceTests: XCTestCase {
```
rodrigoareis, line 15: We need to include some unit tests to test the default and a custom `DeviceIdentifier`.

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+    /// RSA Key types enumeration
+    ///
+    /// - privateKey: Private Key for RSA Key Pair
+    /// - publicKey: Public Key for RSA Key Pair
+    enum DeviceIdentifierKeyType {
+        case privateKey
+        case publicKey
+    }
+    /// Constant Device Identifier Key
+    let deviceIdentifierKey = DeviceIndentifierConstants.deviceIdentifierKey
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: Fixed, I also fixed the build error

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,254 @@
...
+            return deviceIdentifier.id
+        }
+    }
+    
+    
+    /// Hashes given Data using SHA1
+    ///
+    /// - Parameter data: Data to be hashed
+    /// - Returns: Hashed String of given Data
+    /// Legacy method, use `hashAndBase64Data(_:)` instead
```
vahancouver, line 128: Let's update the method name here too

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,254 @@
...
+    /// - Returns: Hashed String of given Data in hexadecimal format
+    /// This is the preferred method to use for hashing device identifiers.
+    func hashSHA256AndTransformToHex(_ data: Data) -> String {
+        let digest = SHA256.hash(data: data)
+        return Data(digest).toHexString()
+    }
+    
+    /// Generates Key Pair, and persists generated Keys
+    ///
+    /// - Returns: A boolean result of whether Key Pair generation, and store process was successful or not
```
vahancouver, line 149: the method doesn't return boolean. the docs need to be updated

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,254 @@
...
+    ///
+    /// - Parameter logger: Optional Logger for logging purposes, default is nil
+    public init(logger: Logger? = nil) {
+        self.logger = logger
+        self.keychainService = KeychainStorage<DeviceIdentifierImpl>(account: deviceIdentifierKey, encryptor: SecuredKeyEncryptor() ?? NoEncryptor())
+    }
+    
+    /// Generates, or retrieves an identifier.
+    ///
+    /// - Returns: Uniquely generated Identifier as per Keychain Sharing Access Group
```
vahancouver, line 93: The code doesn't actually implement Keychain Sharing Access Groups. This should be updated to something more accurate like: - Returns: Uniquely generated identifier stored securely in the device's keychain

vahancouver (approved):
Approved with some minor documentation comments to be addressed

spetrov (approved):
LGTM

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+    /// - Parameter logger: Optional Logger for logging purposes, default is nil
+    public init(logger: Logger? = nil) {
+        self.logger = logger
+        self.keychainService = KeychainStorage<DeviceIdentifierImpl>(account: deviceIdentifierKey, encryptor: SecuredKeyEncryptor() ?? NoEncryptor())
+    }
+    
+    /// Generates, or retrieves an identifier.
+    ///
+    /// - Returns: Uniquely generated identifier stored securely in the device's keychain
+    @discardableResult func getIdentifier() async throws -> String {
```
witrisna, line 94: Synchronize this function for thread safety

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+    
+    
+    /// Hashes given Data using SHA1
+    ///
+    /// - Parameter data: Data to be hashed
+    /// - Returns: Hashed String of given Data
+    /// Legacy method, use `hashSHA256AndTransformToHex(_:)` instead
+    func hashSHA1AndTransformToHex(_ data: Data) -> String {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: We are not using it at the moment. Left it in with a warning in case we want to change our mind before releasing, as we were discussing some migration use cases

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        }
+        let hexString = Data(bytes: digest, count: digest.count).toHexString()
+        return hexString
+    }
+    
+    /// Hashes given Data using SHA256 and converts it to a hexadecimal string
+    /// - Parameter data: Data to be hashed
+    /// - Returns: Hashed String of given Data in hexadecimal format
+    /// This is the preferred method to use for hashing device identifiers.
+    func hashSHA256AndTransformToHex(_ data: Data) -> String {
```
rodrigoareis, line 142: I think this method doesn't interact with shared state I believe could be be marked as `nonisolated` to improve performance and make it clear it doesn't require isolation

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        let digest = SHA256.hash(data: data)
+        return Data(digest).toHexString()
+    }
+    
+    /// Generates Key Pair, and persists generated Keys
+    ///
+    /// - Returns: DeviceIdentifierKeyPair containing Private and Public Keys
+    /// - Throws: An error if key generation or retrieval fails
+    func generateKeyPair() throws -> DeviceIdentifierKeyPair {
```
rodrigoareis, line 151: I think this method could potentially block the thread when performing crypto operations. Mark as `async`?

Device/Device/Device.h
```diff
@@ -0,0 +1,21 @@
...
+//  Device.h
+//  Device
+//
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+#import <Foundation/Foundation.h>
```
rodrigoareis, line 11: Consider adding a README.md to this `Device` module to align with the other modules

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: This is an internal method used only from the DefaultDeviceIdentifer implementation. I made the DefaultDeviceIdentifer into an `Actor` so this and anything will be in isolated context. Setting this or the rest suggested methods to `nonisolated` will not have a performance gain here.

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+    /// - privateKey: Private Key for RSA Key Pair
+    /// - publicKey: Public Key for RSA Key Pair
+    enum DeviceIdentifierKeyType {
+        case privateKey
+        case publicKey
+    }
+    /// Constant Device Identifier Key
+    let deviceIdentifierKey = DeviceIdentifierConstants.deviceIdentifierKey
+    /// Constant Public Key tag
+    let publicKeyTag = DeviceIdentifierConstants.publicKeyTag.data(using: .utf8)!
```
rodrigoareis, line 64: avoid force unwrapping

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        }
+        
+        return query
+    }
+}
+
+
+extension Data {
+    /// Converts Data to a hexadecimal string representation.
+    func toHexString() -> String {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: see other comment

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        return map { String(format: "%02hhx", $0) }.joined()
+    }
+}
+
+enum DeviceIdentifierConstants {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: @frozen has no effect on non-public enums

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: The struct now is an actor and there is also some extra async calls added. It should be thread safe

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: Check the new commit. These are internal methods and the way they are used we would not have any performance benefit.

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,255 @@
...
+        
+        return query
+    }
+    
+    
+    /// Builds Dictionary of Keychain operation attributes for retrieving Key based on given Key Type
+    ///
+    /// - Parameter keyType: RSA Key Type whether Public or Private Key
+    /// - Returns: A dictionary of Keychain operation attributes
+    func buildQuery(_ keyType: DeviceIdentifierKeyType) -> [String: Any] {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: same as above

Device/Device/DeviceIdentifier.swift
rodrigoareis: If `DefaultDeviceIdentifier` has been converted to an actor (rather than a struct as I saw in the code), then the assessment is mostly correct, however marking it as `nonisolated` is appropriate and beneficial, even in an actor. But the impact might not be noticeable.

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+                       userInfo: [NSLocalizedDescriptionKey: "Unable to export public key"])
+            throw DeviceIdentifierError.externalRepresentationFailed(err)
+        }
+        let publicKeyData = pubDataRef as Data
+        return (privateKeyData, publicKeyData)
+    }
+}
+
+/// Concrete identifier storing only key pair; `id` is computed.
+public final class DeviceIdentifierImpl: DeviceIdentifier, Codable {
```
rodrigoareis, line 174: It conforms to `DeviceIdentifier` protocol but doesn't match the required signature. The protocol requires `id` to be `async` `throws` but the implementation is synchronous.

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+                let identifier = hashSHA256AndTransformToHex(keyPair.publicKey)
+                let impl = DeviceIdentifierImpl(deviceIdentifierKeyPair: keyPair)
+                try await keychainService.save(item: impl)
+                return identifier
+            } catch {
+                logger?.e("Key pair generation failed", error: error)
+                logger?.i("Falling back to UUID-based identifier")
+                let uuidData = Data(UUID().uuidString.utf8)
+                let identifier = hashSHA256AndTransformToHex(uuidData)
+                let fallbackPair = DeviceIdentifierKeyPair(privateKey: Data(), publicKey: Data())
```
rodrigoareis, line 74: I think empty `Data()` objects might be confusing and could lead to security issues. Should we consider using a more explicit fallback mechanism? @witrisna do we have a fallback in Android?

Device/DeviceTests/DeviceTests.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+
+import XCTest
+@testable import PingDeviceId
+
+final class DeviceTests: XCTestCase {
```
rodrigoareis, line 15: Should we add tests to cover Keychain access failures or encryption failures?

Device/Device/DeviceId.h
```diff
@@ -0,0 +1,21 @@
...
+//  DeviceId.h
+//  DeviceId
+//
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+#import <Foundation/Foundation.h>
```
rodrigoareis, line 11: Should we rename the folder to `DeviceId` for consistent naming?

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/// Default implementation that generates and persists a device identifier.
+public actor DefaultDeviceIdentifier: DeviceIdentifier {
+    /// Keychain storage service for persisting the device identifier.
+    private let keychainService: any Storage<DeviceIdentifierImpl>
+    /// Optional logger for logging events.
+    private let logger: Logger?
+    /// In-memory cache for the computed identifier.
+    private var cachedId: String?
```
rodrigoareis, line 48: Looks like the `cacheId` is never set

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+        try await Task.detached(priority: .userInitiated) {
+            try DefaultDeviceIdentifier.generateKeyPairSync()
+        }.value
+    }
+    
+    /// Synchronous key-pair generation logic.
+    /// - Throws: `DeviceIdentifierError` if key generation fails.
+    /// - Returns: A `DeviceIdentifierKeyPair` containing the private and public keys.
+    private static func generateKeyPairSync() throws -> DeviceIdentifierKeyPair {
+        let (privData, pubData) = try self.generateRSAKeyPairData(
```
vahancouver, line 105: remove self from a static method

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+                try await keychainService.save(item: impl)
+                return identifier
+            } catch {
+                logger?.e("Key pair generation failed", error: error)
+                logger?.i("Falling back to UUID-based identifier")
+                let uuidData = Data(UUID().uuidString.utf8)
+                let identifier = hashSHA256AndTransformToHex(uuidData)
+                let fallbackPair = DeviceIdentifierKeyPair(privateKey: Data(), publicKey: Data())
+                let fallback = DeviceIdentifierImpl(deviceIdentifierKeyPair: fallbackPair)
+                try await keychainService.save(item: fallback)
```
vahancouver, line 76: This logic seems faulty. 
In the fallback case, you're storing a DeviceIdentifierImpl with empty keys, but the id is computed from the public key. This means that you generate an identifier from UUID data, but store empty key pair and later when retrieved, DeviceIdentifierImpl.id computes from empty public key resulting in a different identifier!

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+        return DeviceIdentifierKeyPair(privateKey: privData, publicKey: pubData)
+    }
+    
+    /// Generates an RSA key pair and returns the private and public key data.
+    /// - Parameters:
+    ///  - keySize: The size of the RSA key in bits.
+    ///  - publicTag: The tag for the public key.
+    ///  - privateTag: The tag for the private key.
+    ///  - Throws: `DeviceIdentifierError` if key generation or export fails.
+    ///  - Returns: A tuple containing the private key data and public key data.
```
vahancouver, line 119: I think last two lines should have one space instead of two after ///

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+                logger?.i("Retrieved device identifier from keychain")
+                return stored.id
+            }
+            // Generate and save new identifier
+            do {
+                let keyPair = try await generateKeyPair()
+                let identifier = hashSHA256AndTransformToHex(keyPair.publicKey)
+                let impl = DeviceIdentifierImpl(deviceIdentifierKeyPair: keyPair)
+                try await keychainService.save(item: impl)
+                return identifier
```
vahancouver, line 68: we should cache before returning

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+            } catch {
+                logger?.e("Key pair generation failed", error: error)
+                logger?.i("Falling back to UUID-based identifier")
+                let uuidData = Data(UUID().uuidString.utf8)
+                let identifier = hashSHA256AndTransformToHex(uuidData)
+                let fallbackPair = DeviceIdentifierKeyPair(privateKey: Data(), publicKey: Data())
+                let fallback = DeviceIdentifierImpl(deviceIdentifierKeyPair: fallbackPair)
+                try await keychainService.save(item: fallback)
+                return identifier
```
vahancouver, line 77: we should cache before returning

vahancouver (changes requested):
Left comments

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: Not sure I follow the train of thought here. The `DeviceIdentifier` protocol defines the `var id: String { get async throws }` property. The intermediate `DeviceIdentifierImpl` conforms to the protocol based on that computed `var` and the compiler allows this. The `getter` implemented can throw and be async if needed, this is useful for the actual `DefaultDeviceIdentifier` that computes this on an `async` and `throwable` way. If that was an issue the compiler would block the build of both this file and also the simpler custom implementation used in the tests.

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: Following up on that:
There’s essentially no upside to adding `nonisolated` on a private actor method:

Isolation isn’t exposed
A private method on an actor can only ever be called from inside that actor, where you already have the actor’s isolation context. You don’t get any extra “freedom” by opting it out of isolation, because no external code can see or call it anyway.

No await / hop savings
Within an actor’s own implementation, calling one isolated (default) method from another doesn’t incur a hop or require await. The compiler knows you’re on the actor’s executor, so you get “free” synchronous calls. Making it nonisolated doesn’t speed anything up.

Additional risk
By marking a private method nonisolated, you remove the compiler’s guarantees that it can’t touch actor-protected state. With no real benefit, you’re trading safety for nothing.

This specific method `hashSHA256AndTransformToHex` is now moved in the protocol as an extension (default implementation) and so is run in a "non isolated" context by default. Adding the `nonisolated` would make no difference same as in the private actor functions.

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: good catch

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+            account: Constants.deviceIdentifierKey,
+            encryptor: SecuredKeyEncryptor() ?? NoEncryptor()
+        )
+    }
+        
+    /// Asynchronously generates a key pair on a background task.
+    /// - Throws: `DeviceIdentifierError` if key generation fails.
+    /// - Returns: A `DeviceIdentifierKeyPair` containing the private and public keys.
+    private func generateKeyPair() async throws -> DeviceIdentifierKeyPair {
+        try await Task.detached(priority: .userInitiated) {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: This is needed to make this `async` in a `userInitiated` queue. Refer to older comment on non blocking threads. In iOS these jobs have next to zero impact nowdays, but we are being extra safe on the existing implementation on where this will run

Device/Device/DeviceIdentifier.swift
george-bafaloukas-forgerock: True this got a bit messed up with the latest changes. I am re-implementing this logic

Device/Device/DeviceIdentifier.swift
```diff
@@ -0,0 +1,203 @@
...
+            }
+            // Generate and save new identifier
+            do {
+                let keyPair = try await generateKeyPair()
+                let identifier = hashSHA256AndTransformToHex(keyPair.publicKey)
+                let impl = DeviceIdentifierImpl(deviceIdentifierKeyPair: keyPair)
+                try await keychainService.save(item: impl)
+                return identifier
+            } catch {
+                logger?.e("Key pair generation failed", error: error)
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: This has been changed and it will throw. We do offer a UUID fallback but this needs to be handled by the developer. 
This makes it cleaner and more transparent instead of downgrading the key without knowing in case of failures. I suggest we do the same in Android

Device/Device/DeviceIdentifier.swift
rodrigoareis: I agree, with the changes you made adding the `nonisolated` to this method would be redundant. You've already correctly marked the `toHexString()` extension method as `nonisolated`. Moving this method to the protocol extension was a good decision.

Device/Device/DeviceIdentifier.swift
rodrigoareis: Any objection on introduce this fallback @witrisna ?

PR #67, opened 2025-08-01
Fix ConfirmationCallback and add unit tests

vahancouver:
[SDKS-4281](https://pingidentity.atlassian.net/browse/SDKS-4281) [iOS][Ping SDK] Polling Wait node with "Exit" option not handled properly

Fix ConfirmationCallback and add unit tests


rodrigoareis (approved):
Changes looks good to me

PR #69, opened 2025-08-06
Sdks 4240 callbacks unit tests

vahancouver:
[SDKS-4240](https://pingidentity.atlassian.net/browse/SDKS-4240)[iOS] Develop Unit Tests for Core Callbacks for Journey Module

Add callback unit tests
Update readme file


rodrigoareis (approved):
Changes looks good to me

Journey/JourneyTests/Callbacks/PasswordCallbackTests.swift
```diff
@@ -0,0 +1,298 @@
...
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import XCTest
+@testable import PingJourney
+
+class PasswordCallbackTests: XCTestCase {
```
george-bafaloukas-forgerock, line 14: It would be a good idea to test that the value is cleared after submission. Same for the ValidatedPasswordCallback

Journey/JourneyTests/Callbacks/AbstractValidatedCallbackTests.swift
```diff
@@ -0,0 +1,322 @@
...
+                    ]
+                  }
+                ],
+                "conditionalPolicies": null
+              }
+            },
+            {
+              "name": "failedPolicies",
+              "value": [
+                "{ \\"params\\": { \\"minLength\\": 3 }, \\"policyRequirement\\": \\"MIN_LENGTH\\" }"
```
george-bafaloukas-forgerock, line 98: It would be good to add multiple failed policies for the tests.

george-bafaloukas-forgerock (approved):
Overall great and we have good coverage. Added a few comments as minor suggestions.

Now that the callbackRegistry is a class it will be good test the concurrency as well. Might be over the top but you could add something like the following

```
func testConcurrentRegistration() {
    let registry = CallbackRegistry()
    let expectation = self.expectation(description: "Concurrent registration completes")
    let iterations = 100 // Register 100 callbacks

    DispatchQueue.global().async {
        DispatchQueue.concurrentPerform(iterations: iterations) { index in
            let key = "callback_\(index)"
            registry.register(type: key, callback: CustomCallback.self)
        }
        expectation.fulfill()
    }

    waitForExpectations(timeout: 5.0)
    // The final count should be exactly 100 if the registry is thread-safe
    XCTAssertEqual(registry.callbacks.count, iterations)
}
```

spetrov (approved):
LGTM!

witrisna (approved):

PR #70, opened 2025-08-07
Add protect callback unit tests and update readme file

vahancouver:
[SDKS-4241](https://pingidentity.atlassian.net/browse/SDKS-4241) [iOS] Implement UNIT TESTS for PingOneProtectInitializationCallback and PingOneProtectEvaluationCallback

Add protect callback unit tests and update readme file

### DIFF
--- a/Protect/ProtectTests/PingOneProtectEvaluationCallbackTests.swift
+++ b/Protect/ProtectTests/PingOneProtectEvaluationCallbackTests.swift
@@ -1,0 +1,368 @@
+//
+//  PingOneProtectEvaluationCallbackTests.swift
+//  ProtectTests
+//
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import XCTest
+@testable import PingProtect
+@testable import PingJourney
+
+// Mock Protect SDK for testing
+class MockProtect {
+    static var shouldThrowError = false
+    static var errorMessage = "Collect data failed"
+    static var pauseBehavioralDataCalled = false
+    static var dataReturnValue = "deviceSignals"
+
+    static func reset() {
+        shouldThrowError = false
+        errorMessage = "Collect data failed"
+        pauseBehavioralDataCalled = false
+        dataReturnValue = "deviceSignals"
+    }
+
+    static func data() async throws -> String {
+        if shouldThrowError {
+            throw TestError.collectDataFailed(errorMessage)
+        }
+        return dataReturnValue
+    }
+
+    static func pauseBehavioralData() async throws {
+        pauseBehavioralDataCalled = true
+    }
+
+    static func resumeBehavioralData() {
+        // Mock implementation
+    }
+
+    static func config(_ config: Any) {
+        // Mock implementation
+    }
+
+    static func initialize() async throws {
+        // Mock implementation
+    }
+}
+
+enum TestError: LocalizedError {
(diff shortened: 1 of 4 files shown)


Protect/ProtectTests/PingOneProtectEvaluationCallbackTests.swift
```diff
@@ -0,0 +1,368 @@
...
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import XCTest
+@testable import PingProtect
+@testable import PingJourney
+
+// Mock Protect SDK for testing
+class MockProtect {
+    static var shouldThrowError = false
```
george-bafaloukas-forgerock, line 17: Consolidate the Mock class in one that covers both use cases and remove duplicated code.

```
// Create a shared test utilities file
protocol MockProtectProtocol {
    static func reset()
    static var shouldThrowError: Bool { get set }
    static var errorMessage: String { get set }
}

class ProtectTestUtilities {
    static func setupBasicCallback<T: AbstractProtectCallback>(_ callback: T, envId: String = "test-env") {
        // Common setup logic
    }
}
```

Protect/ProtectTests/PingOneProtectEvaluationCallbackTests.swift
```diff
@@ -0,0 +1,368 @@
...
+        MockProtect.errorMessage = ""
+
+        // When
+        let result = await callback.collect()
+
+        // Then
+        switch result {
+        case .success:
+            XCTFail("Expected failure")
+        case .failure(let error):
```
george-bafaloukas-forgerock, line 294: Maybe test that the client error message matches what we expect

```
    switch result {
    case .success:
        XCTFail("Expected failure")
    case .failure(let error):
        XCTAssertEqual(error.localizedDescription, JourneyConstants.clientError)
    }
```

Protect/ProtectTests/PingOneProtectInitializeCallbackTests.swift
```diff
@@ -0,0 +1,511 @@
...
+            XCTAssertFalse(MockProtectSDK.resumeBehavioralDataCalled)
+            XCTAssertTrue(MockProtectSDK.pauseBehavioralDataCalled)
+        case .failure:
+            XCTFail("Expected success")
+        }
+    }
+
+    // MARK: - Thread Safety Tests
+
+    func testConcurrentStartCalls() async {
```
george-bafaloukas-forgerock, line 430: Improvement suggestion for the concurrent test:

```
func testConcurrentStartCalls() async {
    callback.initValue(name: JourneyConstants.envId, value: "concurrent-env")
    
    let concurrentCount = 100 // <-- Maybe we can change that to a smaller number
    let tasks = (0..<concurrentCount).map { _ in
        Task { await callback.start() }
    }
    
    let results = await withTaskGroup(of: Result<Void, Error>.self) { group in
        for task in tasks {
            group.addTask { await task.value }
        }
        
        var results: [Result<Void, Error>] = []
        for await result in group {
            results.append(result)
        }
        return results
    }
    
    // Verify all results and check for race conditions
    XCTAssertEqual(results.count, concurrentCount)
}
```

george-bafaloukas-forgerock (changes requested):
Added a few improvement suggestions

Protect/ProtectTests/PingOneProtectEvaluationCallbackTests.swift
vahancouver: Extracted to a separate file, no more duplicates

PR #79, opened 2025-09-25
SDKS-4104 Implement MFA Commons module

rodrigoareis:
# JIRA Ticket

[SDKS-4104](https://pingidentity.atlassian.net/browse/SDKS-4104) Implement MFA Commons module

# Description

The MFA Commons module provides the core foundation and shared functionality for all Multi-Factor Authentication (MFA) modules within the Ping Identity iOS SDK. It includes base implementations for MFA Policies and other common utilities that are leveraged by specialized MFA modules like OATH, Push, and FIDO2.

# Checklist:


MfaCommons/MfaCommons/Util/JwtUtils.swift
```diff
@@ -0,0 +1,197 @@
...
+
+    // MARK: - JWT Validation
+
+    /// Checks if a string is a valid JWT and contains the required fields in its payload.
+    ///
+    /// - Parameters:
+    ///   - jwt: The JWT string to validate.
+    ///   - requiredFields: An array of field names to check in the payload.
+    /// - Returns: `true` if the JWT is valid and contains all required fields, `false` otherwise.
+    public static func isValidJwt(_ jwt: String, requiredFields: [String] = []) -> Bool {
```
george-bafaloukas-forgerock, line 80: This naming is a bit misleading given that we do not check the signature of the JWT. We can either rename it to something like `canParseJwtPayload` or add some validation for the signature

MfaCommons/MfaCommons/URI/UriParser.swift
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    // MARK: - Base64 Utilities
+
+    /// Checks if a string is Base64 encoded.
+    ///
+    /// - Parameter value: The string to check.
+    /// - Returns: `true` if the string is Base64 encoded, `false` otherwise.
+    public func isBase64Encoded(_ value: String) -> Bool {
```
george-bafaloukas-forgerock, line 79: I am not sure the Base64 Utilities should be `Public`. Is there a reason to expose those?

MfaCommons/MfaCommons/Util/JwtUtils.swift
george-bafaloukas-forgerock: Something lilke that:
```
// 5. Verify the signature using CryptoKit
    let isValid = HMAC<SHA256>.isValidAuthenticationCode(signatureData, for: signedData, using: key)

    guard isValid else {
        throw JwtError.invalidSignature
    }
```

george-bafaloukas-forgerock (changes requested):
Looks great Rodrigo. 1 suggestion around the validation and removing some public APIs unless absolutely nessesary.

george-bafaloukas-forgerock:
You need to resolve conflicts as well, as this is far behind develop currently

MfaCommons/MfaCommons/URI/UriParser.swift
rodrigoareis: Yes, this will used by both `OAuth` and `Push` modules.

vahancouver (approved):
Looks good overall

vahancouver (approved):
Approved, but I noticed that the commits are not signed/verified. we might want to correct those

george-bafaloukas-forgerock (approved):
LGTM

PR #93, opened 2025-11-07
SDKS-4105 Implement Push functionality in the MFA Module

rodrigoareis:
# JIRA Ticket

[SDKS-4105](https://pingidentity.atlassian.net/browse/SDKS-4105) Implement Push functionality in the MFA Module

# Description

This change implements Push functionality in the new MFA Module, so that users can receive push notifications for authentication within the Unified SDK.

# Checklist:

### DIFF
--- a/Push/Push/Core/PushService.swift
+++ b/Push/Push/Core/PushService.swift
@@ -1,0 +1,1101 @@
+//
+//  PushService.swift
+//  PingPush
+//
+//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import Foundation
+import PingLogger
+import PingMfaCommons
+import PingOrchestrate
+#if canImport(UIKit)
+import UIKit
+#endif
+
+/// Internal service actor that encapsulates Push MFA business logic.
+///
+/// Responsibilities include credential lifecycle management, notification processing,
+/// device token synchronization, and policy enforcement. The public `PushClient`
+/// exposes a subset of these capabilities to consumers.
+actor PushService {
+
+    // MARK: - Internal Types
+
+    private enum Constants {
+        static let registrationFailureMessage = "Failed to register push credential"
+        static let deviceTokenNotSetMessage = "Device token has not been set"
+    }
+
+    private enum HandlerKeys {
+        static let deviceName = "deviceName"
+    }
+
+    private enum NotificationKeys {
+        static let credentialId = "credentialId"
+        static let messageId = "messageId"
+        static let messageText = "messageText"
+        static let customPayload = "customPayload"
+        static let challenge = "challenge"
+        static let numbersChallenge = "numbersChallenge"
+        static let loadBalancer = "amlbCookie"
+        static let contextInfo = "contextInfo"
+        static let ttl = "ttl"
+        static let pushType = "pushType"
+        static let timeInterval = "timeInterval"
+        static let userId = "userId"
+        static let additionalData = "additionalData"
+    }
+
+    private enum NotificationDefaults {
+        static let ttl = 120
+        static let deviceName = "iOS Device"
+    }
+
+    // MARK: - Stored Properties
+
+    private let storage: any PushStorage
+    private let configuration: PushConfiguration
+    private let httpClient: HttpClient
+    private let policyEvaluator: MfaPolicyEvaluator
+    private let logger: Logger?
+    private let deviceTokenManager: PushDeviceTokenManager
+    private let enableCredentialCache: Bool
+
+    /// In-memory cache for credentials when caching is enabled.
+    private var credentialsCache: [String: PushCredential] = [:]
+
+    /// Dispatch table of push handlers keyed by platform identifier.
+    private let pushHandlers: [String: PushHandler]
+
+    // MARK: - Initialization
+
+    /// Creates a new instance of `PushService`.
+    ///
+    /// - Parameters:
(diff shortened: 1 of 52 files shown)


spetrov (approved):
LGTM!

vahancouver (changes requested):
Minor comments and suggestions

Push/Push/Models/PushCredential.swift
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// ## Example
+    ///
+    /// ```swift
+    /// let uri = "pushauth://push/MyCompany:user@example.com?r=https://...&s=secret..."
+    /// let credential = try await PushCredential.fromUri(uri)
+    /// ```
+    public static func fromUri(_ uri: String) async throws -> PushCredential {
+        // Note: PushUriParser will be implemented in Phase 3
+        throw PushError.invalidUri("URI parsing not yet implemented")
```
george-bafaloukas-forgerock: PushURIParser seems to be implemented in the PR, should  this be used here?

Push/Push/Models/PushCredential.swift
```diff
@@ -0,0 +1,316 @@
...
+    /// - Returns: A URI string representation of this credential.
+    /// - Throws: `PushError.uriFormatting` if formatting fails.
+    ///
+    /// ## Example
+    ///
+    /// ```swift
+    /// let uri = try await credential.toUri()
+    /// // Use uri to generate QR code for backup
+    /// ```
+    public func toUri() async throws -> String {
```
george-bafaloukas-forgerock, line 225: Same as above, should these be like this:
```
public static func fromUri(_ uri: String) async throws -> PushCredential {
    return try await PushUriParser.parse(uri)
}

public func toUri() async throws -> String {
    return try await PushUriParser.format(self)
}
```

Push/Push/Storage/PushKeychainStorage.swift
```diff
@@ -0,0 +1,810 @@
...
+    private let logger: Logger?
+
+    /// The keychain access group for shared keychain access (optional).
+    private let accessGroup: String?
+
+    /// The keychain accessibility level for stored items.
+    private let accessibility: CFString
+
+    /// Queue for serializing keychain operations to ensure thread safety.
+    private let keychainQueue = DispatchQueue(label: "com.pingidentity.push.keychain", qos: .userInitiated)
```
george-bafaloukas-forgerock: This seems like is not used anywhere

Push/Push/Core/PushService.swift
```diff
@@ -0,0 +1,1114 @@
...
+    private let policyEvaluator: MfaPolicyEvaluator
+    private let logger: Logger?
+    private let deviceTokenManager: PushDeviceTokenManager
+    private let enableCredentialCache: Bool
+
+    /// In-memory cache for credentials when caching is enabled.
+    private var credentialsCache: [String: PushCredential] = [:]
+
+    /// Internal registry storing handlers.
+    nonisolated(unsafe) private var handlerRegistry = HandlerRegistry(map: [:])
```
george-bafaloukas-forgerock: This can potentially cause runtime problems. While `HandlerRegistry` is a struct, if pushHandlers is accessed or modified from different isolation contexts (even if primarily initialized in init), it defeats the purpose of the Actor's protection.

Unless you have a specific reason to bypass actor isolation, this variable should just be a standard private var.

george-bafaloukas-forgerock: Ensure the module is on Swift 6 version. If the compiler has issues after this let's catch up and see if we can find a way to make this a bit safer

Push/Push/PushClient.swift
```diff
@@ -0,0 +1,462 @@
...
+    /// - Parameter messageData: Raw notification payload.
+    /// - Returns: The stored ``PushNotification`` (or `nil` when the payload is unsupported).
+    /// - Throws: `PushError` when parsing or persistence fails.
+    public func processNotification(messageData: [String: Any]) async throws -> PushNotification? {
+        try checkInitialized()
+
+        let payload = UnsafeMessagePayload(value: messageData)
+        let notification = try await pushService.processNotification(messageData: payload.value)
+        if let notification {
+            await runAutoCleanup(credentialId: notification.credentialId)
```
george-bafaloukas-forgerock: Should we move this in a Task?

Push/Push/Utilities/PushUriParser.swift
```diff
@@ -0,0 +1,478 @@
...
+    }
+    
+    /// Extract the base server endpoint from a full URL.
+    /// Removes query parameters and _action suffix.
+    ///
+    /// - Parameter url: The full URL with potential query parameters.
+    /// - Returns: The base server endpoint URL without query parameters.
+    /// - Throws: `PushError.invalidUri` if the URL is malformed.
+    private static func extractServerEndpoint(_ url: String) throws -> String {
+        guard let urlComponents = URLComponents(string: url) else {
```
george-bafaloukas-forgerock: There is a chance this is unstable for certain QR codes. This is a minor but we can add a safer way for iOS 17+
```
if let components = URLComponents(string: uri, encodingInvalidCharacters: true) {
    print("Scheme: \(components.scheme ?? "")")
} else {
    print("Failed to parse URI")
}
```

george-bafaloukas-forgerock (changes requested):
Overall looks great! I like the move to MFA Commons for the base32 and the rest.
Have a look on my comments and let me know if you got any questions

george-bafaloukas-forgerock (approved):
LGTM

vahancouver (approved):
LGTM

PR #99, opened 2025-11-18
SDKS-4495-binding migration

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4495) "Sdks 4495 binding migration"

# Description

# Checklist:


rodrigoareis (approved):
Changes looks good to me. Just a missing doc, approving it as log it's addressed before the merge.

Binding/Binding/Config/DeviceBindingConfig.swift
```diff
@@ -20,7 +20,25 @@ public class DeviceBindingConfig {
     #if canImport(UIKit)
     /// The name of the device.
     /// The default is the current device name.
-    public var deviceName: String = UIDevice.current.name
+    private var _deviceName: String?
+    
+    public var deviceName: String {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: Fixed

Binding/Binding/Config/DeviceBindingConfig.swift
```diff
@@ -18,9 +18,27 @@ import UIKit
...
+        get {
+            if let name = _deviceName {
+                return name
+            }
+            if Thread.isMainThread {
+                return UIDevice.current.name
+            } else {
+                return DispatchQueue.main.sync {
+                    UIDevice.current.name
```
witrisna, line 34: Just curious, we can only get the name in main thread?

Binding/Binding/Module/BindingModule.swift
```diff
@@ -29,6 +55,42 @@ public class BindingModule: NSObject {
...
+    
+    /// Triggers the migration check if it hasn't been done yet.
+    /// This is called automatically during module initialization.
+    private static func triggerMigrationIfNeeded() async {
+        let isTriggered = await migrationState.isTriggered()
+        guard !isTriggered else {
+            return
+        }
+        
+        await migrationState.markTriggered()
```
witrisna: Looks like this try to avoid multiple thread or multiple execution of triggerMigrationIfNeeded function.
However, this may not threadsafe. Make function `triggerMigrationIfNeeded` synchronized?

Instead of using MigrationState, can we just use a boolean?

Binding/Binding/Migration/BindingMigration.swift
```diff
@@ -0,0 +1,262 @@
...
+    ///         logger: Logger.standard
+    ///     )
+    ///     print("Migration completed successfully")
+    /// } catch MigrationError.noLegacyDataFound {
+    ///     print("No legacy data to migrate")
+    /// } catch {
+    ///     print("Migration failed: \(error)")
+    /// }
+    /// ```
+    public static func migrate(
```
witrisna, line 123: may need to synchroized migrate()? what if multiple threads execute migrate at the same time?

Binding/Binding/Migration/BindingMigration.swift
```diff
@@ -0,0 +1,262 @@
...
+        accessGroup: String? = nil,
+        logger: Logger? = nil,
+        cleanupLegacyData: Bool = true,
+        storageConfig: UserKeyStorageConfig? = nil
+    ) async throws {
+        // Check if migration has already been attempted
+        let isAttempted = await stateManager.isAttempted()
```
witrisna, line 130: Instead o using stateManager, what about just a boolean isAttempted static in class level?

witrisna (changes requested):

Binding/Binding/Config/DeviceBindingConfig.swift
george-bafaloukas-forgerock: New swift 6 Actor isolation makes some weird changes on this. In iOS anything that is considered UI thread needs to be run in the Main thread. Also the compiler throws build issues if Thread Isolation through the actors is not taken care of.

Binding/Binding/Migration/BindingMigration.swift
george-bafaloukas-forgerock: I am more in favor of the stateManager approach, as this is definitely cover multiple uses of the BindingModule. Also this can be extended to cover other migrations and is thread safe

Binding/Binding/Module/BindingModule.swift
george-bafaloukas-forgerock: Refer to my comment below on the use of StateManager

Binding/Binding/Migration/BindingMigration.swift
george-bafaloukas-forgerock: Addressed

Binding/Binding/Module/BindingModule.swift
george-bafaloukas-forgerock: Latest changes address the threading issue as well

Binding/Binding/Authenticator/PinCollector.swift
```diff
@@ -17,7 +17,7 @@ public protocol PinCollector: AnyObject {
 
 #if canImport(UIKit)
 /// A default implementation of `PinCollector` that uses a `UIAlertController` to prompt the user for their PIN.
-public class DefaultPinCollector: NSObject, PinCollector {
+public class DefaultPinCollector: NSObject, PinCollector, @unchecked Sendable {
```
vahancouver, line 20: Should we mark this @MainActor instead of @unchecked Sendable? It's UI related and should be executed on the Main thread

Binding/Binding/Authenticator/PinCollector.swift
```diff
@@ -27,7 +27,11 @@ public class DefaultPinCollector: NSObject, PinCollector {
     ///   - completion: The closure to call with the collected PIN.
     public func collectPin(prompt: Prompt, completion: @escaping @Sendable (String?) -> Void) {
         
-        DispatchQueue.main.async {
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else {
+                completion(nil)
+                return
+            }
```
vahancouver, line 34: This part might not be needed if we change the class to @MainActor

Binding/PingBindingTests/PingBinderTests.swift
```diff
@@ -80,34 +69,24 @@ class PingBinderTests: XCTestCase {
         let signData = signJsonString.toDictionary()!
         let signCallback = DeviceSigningVerifierCallback().initialize(with: data) as! DeviceSigningVerifierCallback
```
vahancouver: did you mean to use _signData_ instead of _data_?

vahancouver (approved):
Looks good overall. Some minor suggestions

witrisna (approved):

spetrov (approved):
LGTM

PR #103, opened 2025-11-24
Sdks 4491 device client

vahancouver:
[SDKS-4491](https://pingidentity.atlassian.net/browse/SDKS-4491) [iOS] Device Management

Add DeviceClient module
- Source files
- README.md
- Unit Tests
- Podfile
- Update Package.Swift


DeviceClient/DeviceClient/DeviceClient.swift
```diff
@@ -0,0 +1,578 @@
...
+    /// Supported operations:
+    /// - `get()`: Retrieve all Oath devices for the user
+    /// - `delete(_:)`: Delete a specific Oath device
+    ///
+    /// Example:
+    /// ```swift
+    /// let devices = try await client.oath.get()
+    /// try await client.oath.delete(devices.first!)
+    /// ```
+    public lazy var oath: any ImmutableDevice<OathDevice> = {
```
george-bafaloukas-forgerock: Potential retain cycle with the lazy vars here.
The DeviceClient class holds a property (e.g., oath) which stores the ImmutableDeviceImplementation struct. That struct holds a strong reference back to DeviceClient. Because the struct is stored in the class instance, this creates a generic Reference Cycle, preventing DeviceClient from ever being deallocated.

george-bafaloukas-forgerock: Potential fix:

Change the deviceClient property in both implementation structs to unowned. This breaks the cycle by telling the compiler that the struct does not "own" the client, but rather the client owns the struct.

DeviceClient/DeviceClient/Device.swift
```diff
@@ -0,0 +1,254 @@
...
+    /// The ID of the device.
+    public let id: String
+    /// The name of the device.
+    public let deviceName: String
+    /// The URL suffix for the device.
+    public let urlSuffix: String
+    /// The UUID of the device.
+    public let uuid: String
+    /// The creation date of the device.
+    public let createdDate: TimeInterval
```
george-bafaloukas-forgerock, line 110: **Date Handling (Milliseconds vs Seconds)**

You are decoding createdDate and lastAccessDate as TimeInterval (Double). Based on DeviceClientTests.swift (e.g., 1640000000000.0), these are milliseconds since 1970.

iOS Date(timeIntervalSince1970:) expects seconds. If you use these values directly to create Date objects in the UI (as seen in DeviceManagementView.swift), the dates will be approximately 30,000 years in the future.

Potential Fix: Perform the conversion during decoding or provide a computed property.

DeviceClient/DeviceClient/DeviceClient.swift
```diff
@@ -0,0 +1,578 @@
...
+        }
+        
+        do {
+            let json = try response.json()
+            guard let resultArray = json["result"] as? [[String: Any]] else {
+                throw DeviceError.invalidResponse(message: "Missing 'result' array in response")
+            }
+            
+            let resultData = try JSONSerialization.data(withJSONObject: resultArray, options: [])
+            let devices = try JSONDecoder().decode([T].self, from: resultData)
```
george-bafaloukas-forgerock: You can decode directly from the raw response data, no need to turn to data and then decode.

Use a wrapper struct to decode directly like this:

```
private struct DeviceResponse<T: Decodable>: Decodable {
    let result: [T]
}
```

```
do {
        // Decode directly from data
        let wrapper = try JSONDecoder().decode(DeviceResponse<T>.self, from: response.data)
        return wrapper.result
    } catch {
        // ... error handling
    }
```

DeviceClient/DeviceClient/CodableExtentions.swift
```diff
@@ -0,0 +1,258 @@
+// 
+//  CodableExtentions.swift
```
george-bafaloukas-forgerock: Typo this should be: `CodableExtensions`

Package.swift
```diff
@@ -38,5 +39,6 @@ let package = Package (
         .target(name: "PingExternalIdPGoogle", dependencies: [.target(name: "PingExternalIdP"), .product(name: "GoogleSignIn", package: "GoogleSignIn-iOS")], path: "ExternalIdPGoogle/ExternalIdPGoogle", exclude: ["ExternalIdPGoogle.h"], resources: [.copy("PrivacyInfo.xcprivacy")]),
     	.target(name: "PingExternalIdPFacebook", dependencies: [.target(name: "PingExternalIdP"), .product(name: "FacebookLogin", package: "facebook-ios-sdk")], path: "ExternalIdPFacebook/ExternalIdPFacebook", exclude: ["ExternalIdPFacebook.h"], resources: [.copy("PrivacyInfo.xcprivacy")]),
     	.target(name: "PingProtect", dependencies: [.target(name: "PingDavinci"), .product(name: "PingOneSignals", package: "pingone-signals-sdk-ios")], path: "Protect/Protect", exclude: ["Protect.h"], resources: [.copy("PrivacyInfo.xcprivacy")]),
+        .target(name: "PingDeviceClient", dependencies: [.target(name: "PingOrchestrate"), .target(name: "PingLogger")], path: "DeviceClient/DeviceClient", exclude: ["DeviceClient.h"], resources: [.copy("PrivacyInfo.xcprivacy")]),
```
george-bafaloukas-forgerock: Does that have a dependency to Journey as well?
If so, after the plugins are merged this will need to be changed to PingJourneyPlugin

george-bafaloukas-forgerock (changes requested):

DeviceClient/DeviceClient/DeviceClient.swift
```diff
@@ -0,0 +1,578 @@
...
+
+/// Client for managing user devices
+///
+/// `DeviceClient` provides a type-safe interface for managing various types of authentication
+/// devices registered to a user. It supports both immutable devices (read/delete only) and
+/// mutable devices (read/update/delete).
+///
+/// ## Supported Device Types
+///
+/// ### Immutable (Read/Delete Only)
```
spetrov: So, turns out OATH and PUSH devices are mutable, right?

DeviceClient/README.md
```diff
@@ -0,0 +1,318 @@
...
+// Other device types
+let pushDevices = try await deviceClient.push.get()
+let boundDevices = try await deviceClient.bound.get()
+let profileDevices = try await deviceClient.profile.get()
+let webAuthnDevices = try await deviceClient.webAuthn.get()
+```
+
+### Updating Devices
+
+Only mutable device types (Bound, Profile, WebAuthn) support updates:
```
spetrov: So it turned out that OATH and PUSH devices are also mutable...

DeviceClient/README.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+## Device Types
+
+### Oath Device
+
+```swift
+struct OathDevice: Device {
+    let id: String
+    let deviceName: String
```
spetrov: I see we have `// Mutable` comment for the other types...

DeviceClient/README.md
```diff
@@ -0,0 +1,318 @@
...
+    let lastAccessDate: TimeInterval
+}
+```
+
+### Push Device
+
+```swift
+struct PushDevice: Device {
+    let id: String
+    let deviceName: String
```
spetrov: I see we have `// Mutable` comment for the other types...

spetrov (changes requested):
Looks really good to me! 
But it turns out that OATH and PUSH devices are mutable too, so this will require some changes...

DeviceClient/DeviceClient/DeviceClient.swift
vahancouver: one is a struct and the other ones is a class. no potential retain cycle really

DeviceClient/DeviceClient/Device.swift
vahancouver: updated to convert to seconds during decoding

DeviceClient/DeviceClient/DeviceClient.swift
vahancouver: updated

DeviceClient/DeviceClient/CodableExtentions.swift
vahancouver: fixed

Package.swift
vahancouver: no dependency on Journey

DeviceClient/DeviceClient/DeviceClient.swift
vahancouver: fixed in the updated version

DeviceClient/README.md
vahancouver: fixed in the updated version

DeviceClient/README.md
vahancouver: fixed in the updated version

DeviceClient/README.md
vahancouver: fixed in the updated version

DeviceClient/DeviceClient/DeviceClient.swift
george-bafaloukas-forgerock: According to Gemini this might still be an issue, as the reference from the struct will be a strong one:
**1. Strong Reference Cycle (Memory Leak)**
In DeviceClient.swift, DeviceClient holds the repositories (e.g., oath) via lazy var. The repositories are structs (DeviceRepositoryImplementation), but they hold a strong reference back to the DeviceClient class instance. This creates a retain cycle: DeviceClient → oath (struct) → deviceClient (class) → oath ...

Fix: Since DeviceRepositoryImplementation is a struct, it cannot use weak. However, it can use unowned.

DeviceClient/DeviceClient.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,569 @@
...
+					"DEBUG=1",
+					"$(inherited)",
+				);
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 26.0;
```
george-bafaloukas-forgerock, line 347: Set this to 16

george-bafaloukas-forgerock (changes requested):

george-bafaloukas-forgerock:
All good from my side, just double check the gemini suggestion on the retain cycle...might be wrong so I leave that to you. Fix the conflicts when you can and set the module version to 16.

spetrov (approved):
LGTM! 👍🏻

DeviceClient/DeviceClient/CodableExtensions.swift
```diff
@@ -0,0 +1,258 @@
...
+    }
+    
+    init?(intValue: Int) {
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+//MARK:- Decoding Extensions
+extension KeyedDecodingContainer {
```
rodrigoareis, line 28: I think this file is a good candidate to move to new `commons`. When doing it, lets add some docs.

DeviceClient/DeviceClient/DeviceClient.swift
```diff
@@ -0,0 +1,729 @@
...
+struct Session: Codable {
+    let username: String
+    let universalId: String
+    let realm: String
+    let latestAccessTime: String
+    let maxIdleExpirationTime: String
+    let maxSessionExpirationTime: String
+}
+
+public enum DeviceClientConstants {
```
rodrigoareis, line 697: Missing docs

PingDeviceClient.podspec
```diff
@@ -0,0 +1,39 @@
...
+
+  s.source           = {
+      :git => 'https://github.com/ForgeRock/ping-ios-sdk.git',
+      :tag => s.version.to_s
+  }
+
+  s.module_name   = 'PingDeviceClient'
+  s.swift_versions = ['5.0', '5.1', '6.0']
+
+  s.ios.deployment_target = '13.0'
```
rodrigoareis:
```suggestion
  s.ios.deployment_target = '16.0'
```

rodrigoareis (approved):
Overall implementation looks good me. Some minor comments, but those are not a blocker.

DeviceClient/DeviceClient/Device.swift
```diff
@@ -0,0 +1,337 @@
...
+    associatedtype T
+    
+    /// Retrieves a list of devices.
+    /// - Returns: A Result containing either the list of devices or an error.
+    func get() async -> Result<[T], DeviceError>
+    
+    /// Deletes the specified device.
+    /// - Parameter device: The device to delete.
+    /// - Returns: A Result containing either success (true) or an error.
+    func delete(_ device: T) async -> Result<Bool, DeviceError>
```
vibhorgoswami: In Android, we've made this as `Result<T>` which would be translated to `Result<T, DeviceError>`

DeviceClient/DeviceClient/Device.swift
```diff
@@ -0,0 +1,337 @@
...
+    
+    /// Updates the specified device.
+    /// - Parameter device: The device to update.
+    /// - Returns: A Result containing either success (true) or an error.
+    func update(_ device: T) async -> Result<Bool, DeviceError>
```
vibhorgoswami: Similarly here also, changing it to `Result<T, DeviceError>` as per Android.

vibhorgoswami (changes requested):
A couple of changes to match Android.

george-bafaloukas-forgerock (approved):
LGTM

vibhorgoswami (approved):

PR #114, opened 2026-01-13
SDKS-4630 Enhance DefaultDeviceIdentifier to Reuse Legacy Device Identifier When Available

george-bafaloukas-forgerock:
Adding automatic migration of Legacy identifiers. Preserving old identifier and adding ability to regenerate with new SDK (SHA-256 if regenerated).

**Summary of Changes**
This PR introduces a Legacy Migration mechanism to the DeviceId library. The goal is to ensure that users upgrading from the previous FRAuth SDK to the new PingDeviceId library retain their existing device identifiers, rather than generating new ones upon upgrade.

**Key implementations:**

Legacy Detection: Added logic to check the Keychain for data stored by the old SDK (specifically looking for SHA-1 based identifiers).

**Migration Strategy:**

- If a legacy identifier string exists, it is adopted.
 
- If only the legacy public key exists, the identifier is re-calculated using the legacy algorithm (SHA-1) and adopted.
 
- Once found, the legacy ID is wrapped in the new DeviceIdentifierImpl format and stored, so future lookups are fast.
 
- Logic Updates: Modified DefaultDeviceIdentifier to check for legacy data before generating a new RSA key pair.
 
- Cleanup: Updated regenerateIdentifier to ensure it wipes both the new storage and the old legacy storage to guarantee a fresh start.

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4630) "Enhance DefaultDeviceIdentifier to Reuse Legacy Device Identifier When Available"

# Description

# Checklist:

### DIFF
--- a/DeviceId/DeviceId/DefaultDeviceIdentifier.swift
+++ b/DeviceId/DeviceId/DefaultDeviceIdentifier.swift
@@ -19,6 +19,32 @@ import PingStorage
 /// ```swift
 /// let deviceId = DefaultDeviceIdentifier()
 /// let identifier = try await deviceId.id
+///
+/// // In case of migration from the FR SDK when using a custom Keychain Access Group use the following
+///
+///
+/// func setupDeviceIdentifierWithMigration() async throws -> String {
+///     // If your legacy FRAuth SDK used a custom keychain access group,
+///     // specify it here to enable migration
+///     let configuration = DeviceIdentifierConfiguration(
+///         keySize: DeviceIdentifierConfiguration.default.keySize,
+///         keychainAccount: DeviceIdentifierConfiguration.default.keychainAccount,
+///         useEncryption: DeviceIdentifierConfiguration.default.useEncryption,
+///         legacyKeychainAccessGroup: "com.test" // Set Legacy Keychain Access Group
+///     )
+///
+///     // Initialize with configuration
+///    let deviceIdentifier = try DefaultDeviceIdentifier(
+///         configuration: configuration,
+///         logger: LogManager.standard  // Optional: for debugging migration
+///     )
+///
+///     // First access will trigger migration if legacy identifier exists
+///     let deviceId = try await deviceIdentifier.id
+///     print("Device ID: \(deviceId)")
+///
+///     return deviceId
+/// }
 /// ```
 public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
     /// Configuration for the device identifier
@@ -213,7 +239,10 @@ public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
     /// Attempts to migrate a legacy device identifier from FRAuth SDK format
     /// - Returns: The legacy identifier if found and successfully migrated, otherwise nil
     private func migrateLegacyIdentifier() async throws -> String? {
-        let legacyIdentifier = LegacyDeviceIdentifier(logger: logger)
+        let legacyIdentifier = LegacyDeviceIdentifier(
+            accessGroup: configuration.legacyKeychainAccessGroup,
+            logger: logger
+        )
         
         // First try direct retrieval of the identifier string
         if let legacyId = try await legacyIdentifier.getLegacyIdentifier() {
--- a/DeviceId/DeviceId/DefaultDeviceIdentifier.swift
+++ b/DeviceId/DeviceId/DefaultDeviceIdentifier.swift
@@ -193,9 +193,9 @@ public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
             logger?.d("No legacy identifier found, will generate new one")
         } catch {
             // Migration found a legacy ID but failed to store it in new format
-            // Log the error but continue - user experience is more important than preserving legacy ID
-            logger?.e("Failed to migrate legacy device identifier. Will generate a new one instead.", error: error)
-            // Fall through to generate a new identifier
+            // Log the error and continue with generating a new identifier
+            logger?.e("Failed to migrate legacy device identifier, will generate new one instead", error: error)
+            // Fall through to generation step below
         }
         
         // 3. Keychain is empty and no legacy identifier, so generate a new key pair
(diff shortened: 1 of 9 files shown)


DeviceId/DeviceId/DefaultDeviceIdentifier.swift
```diff
@@ -115,7 +116,21 @@ public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
...
+        // Delete from new storage
         try await keychainService.delete()
+        
+        // Also delete from legacy storage to ensure clean regeneration
+        let legacyStorage = LegacyDeviceIdentifier.createLegacyStorage(logger: logger)
+        try? await legacyStorage.delete()
+        
+        // Delete legacy public key storage as well
+        let legacyPublicKeyStorage = KeychainStorage<Data>(
+            account: "com.forgerock.ios.device-identifier.pubic-key.data",
```
vahancouver: typo in .public.

DeviceId/DeviceId/DefaultDeviceIdentifier.swift
```diff
@@ -132,18 +147,59 @@ public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
             return id
         }
         
-        // 2. Keychain is empty, so generate a new key pair
+        // 2. Check for legacy device identifier before generating new one
+        logger?.i("Checking for legacy device identifier")
+        if let legacyId = try await migrateLegacyIdentifier() {
```
vahancouver: should we catch the error here and continue instead of propagating and failing the whole thing?

DeviceId/DeviceId/LegacyDeviceIdentifier.swift
```diff
@@ -0,0 +1,100 @@
...
+import PingLogger
+import PingStorage
+
+/// Legacy device identifier retrieval from FRAuth SDK format.
+/// This class handles migration from the old FRDeviceIdentifier format to the new DeviceId module format.
+internal actor LegacyDeviceIdentifier {
+    /// Legacy keychain key for the identifier
+    private static let legacyIdentifierKey = "com.forgerock.ios.device-identifier.hash-base64-string-identifier"
+    /// Legacy keychain key for public key data
+    private static let legacyPublicKeyDataKey = "com.forgerock.ios.device-identifier.pubic-key.data"
```
vahancouver: same here .. typo in .pubic.

vahancouver: Also why do we have this string in two places?

DeviceId/DeviceIdTests/LegacyDeviceIdentifierMigrationTests.swift
```diff
@@ -0,0 +1,351 @@
...
+        let secondCallId = try await deviceIdentifier.id
+        XCTAssertEqual(generatedId, secondCallId)
+    }
+    
+    /// Test migration from public key data when identifier doesn't exist but public key does
+    func testMigrationFromLegacyPublicKey() async throws {
+        // GIVEN: Legacy public key data exists but no identifier
+        let samplePublicKeyData = Data("sample_public_key_data".utf8)
+        let legacyPublicKeyStorage = KeychainStorage<Data>(
+            account: "com.forgerock.ios.device-identifier.pubic-key.data",
```
vahancouver: typo in .public.

vahancouver (changes requested):
Left some comments to address

DeviceId/DeviceId/DefaultDeviceIdentifier.swift
george-bafaloukas-forgerock: That is not a "typo". Unfortunately that was how it was on the old SDK, so we need to keep it like this

DeviceId/DeviceId/LegacyDeviceIdentifier.swift
george-bafaloukas-forgerock: yeah let me look on consolidating those. But yeah this is not a typo.

DeviceId/DeviceIdTests/LegacyDeviceIdentifierMigrationTests.swift
george-bafaloukas-forgerock: See other responses please

spetrov (approved):
Looks good to me. The logic makes sense to me, but when I tried an "upgrade" scenario from the legacy SDK, the identifier is not retrieved from the keychain.
Debugging shows somewhat different query being used.

george-bafaloukas-forgerock:
> Looks good to me. The logic makes sense to me, but when I tried an "upgrade" scenario from the legacy SDK, the identifier is not retrieved from the keychain. Debugging shows somewhat different query being used.

Looking into that

DeviceId/DeviceId/DefaultDeviceIdentifier.swift
george-bafaloukas-forgerock: fixed

DeviceId/DeviceId/LegacyDeviceIdentifier.swift
george-bafaloukas-forgerock: Fixed

vahancouver (approved):
LGTM

DeviceId/DeviceId/DefaultDeviceIdentifier.swift
```diff
@@ -106,6 +132,7 @@ public actor DefaultDeviceIdentifier: DeviceIdentifier, Sendable {
     
     /// Asynchronously regenerates the device identifier by deleting the existing keychain item.
     /// This method cancels any ongoing generation task and clears the cache.
+    /// Also clears any legacy identifier storage to ensure a completely new identifier is generated.
     /// - Throws: `DeviceIdentifierError` if keychain operations fail.
     /// - Returns: The new unique identifier for the device.
     public func regenerateIdentifier() async throws -> String {
```
witrisna, line 138: We don't have this in Android, will consider adding it in the future.

witrisna (approved):

spetrov (approved):

PR #115, opened 2026-01-13
Sdks 4604 network module integration

vahancouver:
[SDKS-4604](https://pingidentity.atlassian.net/browse/SDKS-4604) [iOS] Implement Network module 

- Remove Network related code from Orchestrate
- Update Network module to be more aligned with Android (try/catch vs result based, simplified request/response internal impl.)
- Integrate Network module instead of Orchestrate in other modules 
- Update all unit tests, make sure 100% SUCCEED
- Update dependencies according to the new structure
- Update the docs according to the new changes
- Add missing podfiles
- Update Package.swift (reformat for better readability) (with minimal dependencies taking account the transitive dependencies)
- Update podfiles (with minimal dependencies taking account the transitive dependencies)

Here is updated [Ping iOS SDK Module Dependencies](https://pingidentity.atlassian.net/wiki/x/0oFxjQ)


Network/Network/URLSession/URLSessionHttpRequest.swift
```diff
@@ -193,82 +211,126 @@ public final class URLSessionHttpRequest: HttpRequest, @unchecked Sendable {
...
+    // Helper to serialize a JSON dictionary to Data 
+    private func serializeJSON(_ json: [String: Any]) -> Data? {
         if json.isEmpty {
-            return .success(Data("{}".utf8))
+            return Data("{}".utf8)
         }
         
         guard JSONSerialization.isValidJSONObject(json) else {
-            return .failure(NetworkError.invalidRequest("Invalid JSON body"))
+            jsonSerializationFailed = true
```
george-bafaloukas-forgerock, line 270: jsonSerializationFailed = true 
Failure is recorded but error details are lost. Maybe we need to log

Network/Network/URLSession/URLSessionHttpClient.swift
```diff
@@ -191,6 +180,41 @@ public final class URLSessionHttpClient: HttpClientProtocol, @unchecked Sendable
...
+    public func logResponse(responseData: Data?, response: URLResponse?) {
+        var log = "⬇\n"
+        if let httpResponse = response as? HTTPURLResponse {
+            log += "Response Status Code: \(httpResponse.statusCode)\n"
+            log += "Response Headers: \(httpResponse.allHeaderFields)\n"
+            log += "Response Value: \(httpResponse.debugDescription)\n"
+        }
+        
+        if let data =  responseData, let dataString = String(data: data, encoding: .utf8) {
+            log += "Response Data: \(dataString)"
```
george-bafaloukas-forgerock: This logs the entire response body. If the response contains PII or sensitive tokens, this is a security risk. If the response is a 5MB image, it is a performance risk. Improvement:

- Add a generic truncation limit 

- Add a configuration flag to HttpClientConfig to disable body logging for production builds

Network/Network/URLSession/URLSessionHttpRequest.swift
```diff
@@ -17,25 +17,26 @@ import Foundation
 ///
 /// **This class is NOT thread-safe.** It contains mutable state without synchronization
 /// and should not be shared across multiple threads or modified concurrently.
-public final class URLSessionHttpRequest: HttpRequest, @unchecked Sendable {
+public class URLSessionHttpRequest: HttpRequest, @unchecked Sendable {
```
george-bafaloukas-forgerock, line 21: The class is marked as @unchecked Sendable but contains mutable state. 
Since URLRequest is a value type and there's no synchronization, concurrent access could cause issues.
Recommendation: Either make it truly immutable after construction, add proper synchronization, or document that it's not thread-safe during construction.

Network/Network/HttpResponse.swift
```diff
@@ -24,9 +24,6 @@ public protocol HttpResponse: Sendable {
     /// The response body as raw data.
     var body: Data? { get }
 
-    /// All response headers as a dictionary.
-    var headers: [String: [String]] { get }
```
george-bafaloukas-forgerock, line 28: Why have we removed the headers from this?

george-bafaloukas-forgerock (commented):
Looks good overall. As a first pass please see these comments. Additionally since the RequestAPI is changing and that is a breaking change please document a migration guide.

```
// Old API
request.header(name: "X-Custom", value: "value")
request.body(body: ["key": "value"])

// New API
request.setHeader(name: "X-Custom", value: "value")
request.post(json: ["key": "value"])

```

spetrov (approved):
LGTM! Great job! 👍🏻

Network/Network/URLSession/URLSessionHttpRequest.swift
vahancouver: Added Logs for more info

Network/Network/URLSession/URLSessionHttpClient.swift
vahancouver: added internal limit

Network/Network/URLSession/URLSessionHttpRequest.swift
vahancouver: Docs already state that **This class is NOT thread-safe.**

Network/Network/HttpResponse.swift
vahancouver: to keep aligned with Android and it was redundant anyway

george-bafaloukas-forgerock (approved):
Nice!LGTM

Network/NetworkTests/URLSessionHttpResponseTests.swift
```diff
@@ -12,49 +12,13 @@ import XCTest
 @testable import PingNetwork
 
 final class URLSessionHttpResponseTests: XCTestCase {
```
rodrigoareis, line 14: I believe we need to add more tests to it. Specially for the cookie parsing logic.

Network/NetworkTests/URLSessionHttpClientTests.swift
```diff
@@ -2,7 +2,7 @@
 //  URLSessionHttpClientTests.swift
 //  PingNetworkTests
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
```
rodrigoareis, line 5: I would suggest to include a test for actual timeout behavior

rodrigoareis (approved):
The Network module integration is well-implemented with excellent documentation. Just left minor comments.

Network/NetworkTests/URLSessionHttpClientTests.swift
vahancouver: added!

Network/NetworkTests/URLSessionHttpResponseTests.swift
vahancouver: good catch. we definitely needed more unit tests here. added!

PR #117, opened 2026-01-21
SDKS-4717 - Fixing Browser close and reset logic

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4717) "Close External Browser (External IdP Authentication) if user doesn't respond in specified time"

# Description

**Summary of Changes**
The diff represents a refactor of the BrowserLauncher class, specifically focusing on how web authentication sessions are managed and cancelled.

**Consolidated Cancellation Logic**: The private close() method has been removed. Its logic (handling specific session types like SFSafariViewController and ASWebAuthenticationSession) has been moved into the reset() method (inferred name based on context/code structure).

**Fixed Continuation Leak**: In asWebAuthenticationSession, the code now captures the continuation into the instance property self.loginContinuation immediately. Previously, the continuation was only local to the closure, making it impossible for external methods (like a cancel button) to fail the specific active await call.

**Explicit Session Handling**: The reset logic now explicitly checks for ASWebAuthenticationSession to call .cancel(), whereas previously it only specialized for SFSafariViewController and treated everything else generically.

**Logging**: Added info logs (logger.i) to track which type of session is being reset.

# Checklist:

- [] I added test case coverage for my changes

### DIFF
--- a/Browser/Browser/BrowserLauncher.swift
+++ b/Browser/Browser/BrowserLauncher.swift
@@ -302,8 +302,8 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
                     self.state = .closing
                     
                     // Close the UI
-                    safariVC.dismiss(animated: true) {
-                        self.cleanup()
+                    safariVC.dismiss(animated: true) { [weak self] in
+                        self?.cleanup()
                     }
                 }
         }
--- a/Browser/Browser/BrowserLauncher.swift
+++ b/Browser/Browser/BrowserLauncher.swift
@@ -69,7 +69,7 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
     // MARK: Properties
     
     /// Static shared instance
-    public static let currentBrowser: BrowserLauncherProtocol = BrowserLauncher()
+    public static var currentBrowser: BrowserLauncherProtocol = BrowserLauncher()
     
     public var isInProgress: Bool {
         if case .idle = state { return false }
--- a/Browser/Browser/BrowserLauncher.swift
+++ b/Browser/Browser/BrowserLauncher.swift
@@ -127,15 +127,24 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
         
         state = .closing
         
-        if let session = session as? SFSafariViewController {
-            session.dismiss(animated: false) {
+        // Resume continuation with cancellation error
+        loginContinuation?.resume(throwing: BrowserError.externalUserAgentCancelled)
+        
+        // Handle cleanup based on session type
+        if let sfViewController = session as? SFSafariViewController {
+            logger.i("Resetting SFSafariViewController session")
+            sfViewController.dismiss(animated: false) {
                 self.cleanup()
             }
+        } else if let asAuthSession = session as? ASWebAuthenticationSession {
+            logger.i("Resetting ASWebAuthenticationSession")
+            asAuthSession.cancel()
+            cleanup()
         } else {
+            // Native browser or unknown session type
+            logger.i("Resetting non-UI session")
             cleanup()
         }
-        
-        loginContinuation?.resume(throwing: BrowserError.externalUserAgentCancelled)
     }
     
     /// Launches external user-agent for web requests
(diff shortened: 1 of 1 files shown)


Browser/Browser/BrowserLauncher.swift
```diff
@@ -51,91 +52,91 @@ public protocol BrowserLauncherProtocol: Sendable {
...
             logger.i("App became active during native browser authentication. Cancelling.")
-            loginContinuation?.resume(throwing: BrowserError.externalUserAgentCancelled)
-            cleanup()
+            reset()
         }
     }
     
-    /// Resets the browser state and dismisses any presented view controllers.
-    /// - Note: If the browser is not in an authenticating state, this method will log a warning and return without making any changes.
+    /// Resets the browser state, cancels continuations, and dismisses controllers.
```
witrisna, line 100: Looks like developer needs to handle the timeout and call reset. Maybe we can launch the URL with a timeout value, the SDK handle it automatically.
After `launch()` if not completed with the browser, return a timeout error to developer.

george-bafaloukas-forgerock: Yes that is an option. I didn't go too deep on that as it would change the API and went for minimum changes as this can be handled on the UI. If we are in favor of the approach the SDK handling that then I can implement it. What would be the default timeout value in that case? 5 mins?

Browser/Browser/BrowserLauncher.swift
```diff
@@ -51,91 +52,91 @@ public protocol BrowserLauncherProtocol: Sendable {
...
-                self.cleanup()
+        // 1. Capture and Nil the continuation to prevent double-resume
+        let pendingContinuation = self.loginContinuation
+        self.loginContinuation = nil
+        
+        // 2. Resume with cancellation error immediately
+        pendingContinuation?.resume(throwing: BrowserError.externalUserAgentCancelled)
+        
+        // 3. Handle Session Cleanup
+        if let sfViewController = session as? SFSafariViewController {
```
vibhorgoswami, line 125: Are you trying to close the browser here?

george-bafaloukas-forgerock: Yes calling reset() in iOS will close the SafariViewController or the ASWebAuthentication session. We have control over this and can dismiss. We will also clean up any continuations as needed. 

If you have opted in for "NativeBrowser" we cannot interfere on this.

Browser/Browser/BrowserLauncher.swift
```diff
@@ -239,55 +217,98 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
...
+                    // SFVC flow successful
+                    if let activeContinuation = self.loginContinuation {
+                        activeContinuation.resume(returning: url)
+                        self.loginContinuation = nil // Consumed
+                    }
+                    
+                    self.state = .closing
+                    
+                    // Close the UI
+                    safariVC.dismiss(animated: true) {
```
vahancouver: Should we capture self with [weak self] here as well?

vahancouver (approved):
Approved with two comments

Browser/Browser/BrowserLauncher.swift
george-bafaloukas-forgerock: Will update

Browser/Browser/BrowserLauncher.swift
```diff
@@ -297,94 +318,122 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
...
+        return try await withCheckedThrowingContinuation { [weak self] continuation in
+            guard let self = self else {
+                continuation.resume(throwing: BrowserError.externalUserAgentFailure)
+                return
+            }
+            
+            self.loginContinuation = continuation
+            
+            let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: callbackURLScheme) { [weak self] (callbackURL, error) in
                 guard let self = self else { return }
```
vahancouver, line 331: Should we resume continuation here as well?: 
```
continuation.resume(throwing: BrowserError.externalUserAgentFailure)
        return 
    }
```

george-bafaloukas-forgerock: I don't think we do as this returns here in case of failure:

```
if !authSession.start() {
                self.logger.e("Failed to start ASWebAuthenticationSession", error: nil)
                self.state = .closing
                // Resume immediately on failure
                self.loginContinuation?.resume(throwing: BrowserError.externalUserAgentFailure)
                self.loginContinuation = nil
                self.cleanup()
            }
```

george-bafaloukas-forgerock: Truth be told this is redundand anyways cause if self is not resolved the continuation won't be called either ;)

Browser/Browser/BrowserLauncher.swift
vibhorgoswami: Perhaps calling it `close()` instead of reset would be better justified as the browser is being closed.

george-bafaloukas-forgerock: Not sure I agree. As this resets the state of the browser in general and allows for a new authentication/use, not only closes it.

vibhorgoswami (approved):

spetrov (approved):
LGTM!

PR #119, opened 2026-01-26
Sdks 4678  fix unit tests

vahancouver:
[SDKS-4678](https://pingidentity.atlassian.net/browse/SDKS-4678) [Ping SDK] [iOS] Investigate all excluded tests, fix and mark them included

Addressed all excluded tests. No more excluded tests as of now

I'm not super happy about our LocationManager, but maybe we can rewrite it as part of another ticket. the continuation logic is not 100%robust


Binding/PingBindingTests/PingBindingTests.swift
```diff
@@ -469,30 +469,6 @@ final class PingBindingTests: XCTestCase {
         }
     }
     
-    func testUserKeysStorage_ConcurrentAccess() async {
```
george-bafaloukas-forgerock, line 472: Why are we removing this one?

DeviceProfile/DeviceProfileTests/DeviceProfileCollectorTests.swift
```diff
@@ -12,6 +12,7 @@ import XCTest
 import PingLogger
 @testable import PingDeviceProfile
 @testable import PingDeviceId
+import _LocationEssentials
```
george-bafaloukas-forgerock: What is this?

DeviceProfile/DeviceProfileTests/BluetoothCollectorTests.swift
```diff
@@ -183,35 +131,19 @@ class BluetoothCollectorTests: XCTestCase {
...
-    }
-    
-    func testBluetoothCollectorDescription() {
-        // Verify collector can be described without crashing
-        let description = String(describing: collector)
-        XCTAssertFalse(description.isEmpty, "Collector description should not be empty")
-        XCTAssertTrue(description.contains("BluetoothCollector"),
-                     "Description should mention BluetoothCollector")
+    func getBluetoothSupported() async -> Bool {
+        let isBLESupported = mockState == .poweredOn || mockState == .poweredOff
```
george-bafaloukas-forgerock, line 146: In RealBluetoothStateProvider.getBluetoothStatus(), you are using a Stream that yields a value and finishes. Ensure that if the Bluetooth state changes from .unknown to .poweredOn, the collector handles the transition or waits sufficiently, otherwise it might return false prematurely.

george-bafaloukas-forgerock (approved):
LGTM, some minor comments

DeviceProfile/DeviceProfileTests/DeviceProfileCollectorTests.swift
vahancouver: Xcode suggested as subset of CoreLocation. Changed it to import CoreLocation

DeviceProfile/DeviceProfileTests/BluetoothCollectorTests.swift
vahancouver: Fixed to consider that case!

Binding/PingBindingTests/PingBindingTests.swift
vahancouver: Good question! 
So this test **ALWAYS passes on iOS 26** device and **ALWAYS fails on iOS 16** device. which means it is inherently a flaky test. After consulting with copilot 👯  I decided to remove it. To be honest not even sure what it is testing... Seems like we are testing the system not our code, which turns out behaves differently on different iOS versions. We can discuss this over a call if you still think we need this test or a modified version of it. let me know

spetrov (approved):
LGTM! 👍🏻

PR #125, opened 2026-02-13
SDKS-4770 Aligning the storage strategy with Android

george-bafaloukas-forgerock:
Aligning the storage strategy with Android and adding tests. Adding a Coaching guide as well for explaining the way the new cache works

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4770) "Aligning the storage strategy with Android"

# Description

This PR introduces a robust CacheStrategy enum to replace the legacy boolean cacheable flag, providing more granular control over how data is persisted and cached. It also includes comprehensive documentation and updated test suites to reflect these changes.

**Description of Changes**
**1. Core Storage Logic (StorageDelegate.swift & MemoryStorage.swift)**
Introduced CacheStrategy Enum: Replaced the simple Bool toggle with three distinct strategies:

.NO_CACHE: Operations go directly to storage (Default).

.CACHE: Items are cached on save; reads are served from memory first.

.CACHE_ON_FAILURE: Storage is tried first; cache serves as a fallback if storage fails (ideal for network-backed storage).

Actor-Based Caching: The internal CacheManager and the Memory storage are implemented as Swift Actors, ensuring thread-safe access to cached data across concurrent tasks.

Deprecation Path: The old cacheable: Bool initializers are now marked with @available(*, deprecated) to guide developers toward the new enum-based API while maintaining backward compatibility.

**2. Documentation (CACHING_GUIDE.md)**
A massive new guide was added, covering architecture, flow diagrams for each strategy, performance characteristics, and migration examples.

**3. Test Coverage**
Updated Existing Tests: Refactored DaVinciTests and DaVinciErrorTests to explicitly use .NO_CACHE to maintain existing behavior while adopting the new API.

New Strategy Tests: Added comprehensive test cases in StorageDelegateTests.swift to verify the logic of CACHE_ON_FAILURE and concurrent access safety.

# Checklist:

### DIFF
--- a/Storage/Storage/MemoryStorage.swift
+++ b/Storage/Storage/MemoryStorage.swift
@@ -129,7 +129,7 @@ public actor Memory<T: Codable & Sendable>: Storage {
 ///
 /// ## Migration from Legacy API
 ///
-/// The `cacheable` parameter is deprecated. Use `cacheStrategy` instead:
+/// The `cacheable` parameter is removed. Use `cacheStrategy` instead:
 ///
 /// **Before:**
 /// ```swift
@@ -194,32 +194,5 @@ public class MemoryStorage<T: Codable & Sendable>: StorageDelegate<T>, @unchecke
     /// - SeeAlso: `CacheStrategy`, `StorageDelegate`
     public init(cacheStrategy: CacheStrategy = .NO_CACHE) {
         super.init(delegate: Memory<T>(), cacheStrategy: cacheStrategy)
-    }
-    
-    /// Initializes a new instance of `MemoryStorage` with boolean caching flag.
-    ///
-    /// - Warning: This initializer is deprecated. Use `init(cacheStrategy:)` instead.
-    ///
-    /// The `cacheable` parameter is mapped to cache strategies as follows:
-    /// - `true` maps to `.CACHE`
-    /// - `false` maps to `.NO_CACHE`
-    ///
-    /// ## Migration Guide
-    ///
-    /// **Old code:**
-    /// ```swift
-    /// let storage = MemoryStorage<User>(cacheable: true)
-    /// ```
-    ///
-    /// **New code:**
-    /// ```swift
-    /// let storage = MemoryStorage<User>(cacheStrategy: .CACHE)
-    /// ```
-    ///
-    /// - Parameter cacheable: A Boolean value indicating whether the stored data should be cached.
-    ///                        `true` enables caching (equivalent to `.CACHE`), `false` disables it (`.NO_CACHE`).
-    @available(*, deprecated, message: "Use init(cacheStrategy:) instead. This initializer will be removed in a future version.")
-    public init(cacheable: Bool = false) {
-        super.init(delegate: Memory<T>(), cacheable: cacheable)
     }
 }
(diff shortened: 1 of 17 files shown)


spetrov (approved):
LGTM!

Storage/StorageTests/StorageDelegateTests.swift
```diff
@@ -18,7 +18,7 @@ final class StorageDelegateTests: XCTestCase, @unchecked Sendable {
     
     override func setUp() {
         super.setUp()
-        memoryStorage = MemoryStorage()
+        memoryStorage = MemoryStorage(cacheStrategy: .NO_CACHE)
         storageDelegate = StorageDelegate(delegate: memoryStorage, cacheable: false)
```
vahancouver: This needs to be updated as well?!

vahancouver (approved):
LGTM, But are we sure we wanna depreciate instead of breaking changes (which we are allowed to do in this version)? Now we have warnings in the tests that are using those deprecated APIs. What's the practice on the Android side?

Storage/StorageTests/StorageDelegateTests.swift
george-bafaloukas-forgerock: I am generally in favor on deprecating and then removing on the next major version. Apple does that as well, deprecates and removes on future versions. It's easy to rip out the bandage now as we don't have a big use base. Happy to discuss this and we make a team decision on this though

vahancouver: I agree with that. But as we are allowed to have breaking changes and we do have in other places, I thought we could keep it clean of deprecations up to 2.0.0. The ticket I'm working on has two Type name changes which will be a complete nightmare to implement with deprecations. So I'm introducing breaking changes. So I thought I'll bring it up for your feature too.

Storage/Storage/MemoryStorage.swift
```diff
@@ -34,22 +71,154 @@ public actor Memory<T: Codable & Sendable>: Storage {
...
+    ///
+    /// The `cacheable` parameter is mapped to cache strategies as follows:
+    /// - `true` maps to `.CACHE`
+    /// - `false` maps to `.NO_CACHE`
+    ///
+    /// ## Migration Guide
+    ///
+    /// **Old code:**
+    /// ```swift
+    /// let storage = MemoryStorage<User>(cacheable: true)
```
witrisna: Since we are releasing 2.0, breaking change is acceptable. I believe we can remove the deprecated code.

Storage/StorageTests/MemoryStorageTests.swift
```diff
@@ -18,8 +18,8 @@ final class MemoryStorageTests: XCTestCase {
     
     override func setUp() {
         super.setUp()
-        memoryStorage = MemoryStorage()
-        memoryStorageMulti = MemoryStorage()
+        memoryStorage = MemoryStorage(cacheStrategy: .NO_CACHE)
```
witrisna: Can we default to NO_CACHE? so no parameter is not required to construct the object.

witrisna (changes requested):

Storage/StorageTests/MemoryStorageTests.swift
george-bafaloukas-forgerock: If we remove the deprecation yes. I will re-commit the updated one

vahancouver (approved):
LGTM

witrisna (approved):

PR #131, opened 2026-03-06
SDKS-4817-Adding localized error and sendable compliance

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4817) "Adding localized error and sendable compliance"

# Description

**NOTE**: Also addresses [TRIAGE-33163](https://pingidentity.atlassian.net/browse/TRIAGE-33163)

**1. Concurrency Safety (Sendable Adoption)**
The PR adds Sendable conformance to almost every public Error and enum across the workspace (Binding, Browser, Davinci, Fido, etc.).

This allows these error objects to be safely passed between actors and across distributed isolation boundaries in Swift 6/Strict Concurrency mode.

Note on ApiError: This uses @unchecked Sendable for ApiError in the Davinci and Journey modules. This is a pragmatic choice because they wrap [String: Any] (which isn't inherently Sendable), but as noted in the comments, the data is immutable after creation.

**2. Improved Error Localization (LocalizedError)**
Many enums were updated to conform to LocalizedError by implementing errorDescription.

User Experience: This ensures that when a developer accesses error.localizedDescription, they get a human-readable string instead of just the enum case name.

Standardization: It replaces custom property names (like errorMessage) with the standard protocol requirements.

# Checklist:

### DIFF
--- a/Binding/Binding/Errors/DeviceBindingError.swift
+++ b/Binding/Binding/Errors/DeviceBindingError.swift
@@ -2,7 +2,7 @@
 //  DeviceBindingError.swift
 //  PingBinding
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
--- a/Binding/Binding/Errors/DeviceBindingError.swift
+++ b/Binding/Binding/Errors/DeviceBindingError.swift
@@ -11,7 +11,7 @@
 import Foundation
 
 /// An enum representing the possible errors that can occur during the device binding and signing process.
-public enum DeviceBindingError: Error, Equatable {
+public enum DeviceBindingError: Error, LocalizedError, Equatable, Sendable {
     /// Authentication with the user failed.
     case authenticationFailed
     /// The device does not support the required authentication method (e.g., biometrics).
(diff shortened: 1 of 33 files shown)


Binding/Binding/Migration/MigrationError.swift
```diff
@@ -11,7 +11,7 @@
 import Foundation
 
 /// Errors that can occur during device binding migration.
-public enum MigrationError: LocalizedError {
+public enum MigrationError: LocalizedError, Sendable {
     /// Failed to read legacy user keys from keychain
     case failedToReadLegacyKeys(Error)
```
vahancouver, line 16: The associated value type (Error) is not Sendable. I see in other cases like this we marked the enum as @unchecked Sendable (like the ApiError with associated value type [String: Any]). Should we do the same here? Is the compiler not complaining as it is now?

Binding/Binding/Errors/DeviceBindingError.swift
vahancouver, line 24: Please see my comment under MigrationError. same applies here

Storage/Storage/SecuredKeyEncryptor.swift
```diff
@@ -65,4 +65,7 @@ public enum EncryptorError: LocalizedError {
             return "Failed to decrypt given data"
         }
     }
+
+    /// A localized description of the error, used by `LocalizedError`.
+    public var errorDescription: String? { errorMessage }
```
vahancouver: should we just get rid of errorMessage property and use errorDescription instead. Kinda weird to have two properties returning the same value.

vahancouver (approved):
Approved with some questions!

Binding/Binding/Migration/MigrationError.swift
george-bafaloukas-forgerock: From the latest (if not mistaken) version of Swift Error is `Sendable`.
```
public protocol Error : Sendable {
}
```
The problem with certain types is that the use of `Any` would not be sendable and swapping that to other types in some cases is more trouble than it brings benefits

Binding/Binding/Errors/DeviceBindingError.swift
george-bafaloukas-forgerock: See reply above

Storage/Storage/SecuredKeyEncryptor.swift
george-bafaloukas-forgerock: Good idea

spetrov (approved):
LGTM! 👍🏻

PR #132, opened 2026-03-11
Add ReCaptcha Enterprise integration tests

vahancouver:
[SDKS-4379](https://pingidentity.atlassian.net/browse/SDKS-4379) [iOS] Test ReCaptcha Callbacks

1. Add ReCaptcha Enterprise i**ntegration tests**
2. Fix the bug when _non async_ version of **`initialize`** method was being used in **`AbstractProtectCallback`** and **`ReCaptchaEnterpriseCallback`**
3. Remove _non async_ version of **initialize** method from **`Callback`** protocol and **`AbstractCallback`** class as it is never called and has been replaced by _async_ version of **`initialize`** in JourneyPlugin
4. Update all Callback Tests to use _async_ version of **`initialize`** in Journey module tests

### DIFF
--- a/JourneyPlugin/JourneyPlugin/AbstractCallback.swift
+++ b/JourneyPlugin/JourneyPlugin/AbstractCallback.swift
@@ -19,11 +19,6 @@ open class AbstractCallback: Callback, @unchecked Sendable {
 
     /// Initializes a new instance of `AbstractCallback` with the provided JSON.
     open func initialize(with json: [String: Any]) async -> any Callback {
-        return self.privateInit(with: json)
-    }
-    
-    /// Private convinient initializer, to allow both Async and Sync initializations
-    private func privateInit(with json: [String: Any]) -> any Callback {
         self.json = json
         if let output = json[JourneyConstants.output] as? [[String: Any]] {
             for item in output {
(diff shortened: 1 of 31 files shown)


JourneyPlugin/JourneyPlugin/AbstractCallback.swift
```diff
@@ -21,12 +21,6 @@ open class AbstractCallback: Callback, @unchecked Sendable {
...
     }
-
-    
-    /// Initializes a new instance of `AbstractCallback` with the provided JSON.
-    open func initialize(with json: [String: Any]) -> any Callback {
-        return self.privateInit(with: json)
-    }
     
     /// Private convinient initializer, to allow both Async and Sync initializations
     private func privateInit(with json: [String: Any]) -> any Callback {
```
george-bafaloukas-forgerock: Do we need this method now? Maybe we should get the contents in `open func initialize(with json: [String: Any]) async -> any Callback`

george-bafaloukas-forgerock (approved):
One minor improvement suggested

spetrov (approved):
LGTM. 👍🏻 But one of the e2e tests is failing...

PR #137, opened 2026-03-23
Make httpClient public in WorkFlowConfig

vahancouver:
[SDKS-4861](https://pingidentity.atlassian.net/browse/SDKS-4861)

[iOS][Unified SDKs] Allow the setting of RequestInterceptors in Journey and Davinci Workflow

- Make `httpClient` setter public in `HttpClientProtocol`
- Update `RequestInterceptor` docs to better reflect it's purpose

Note: `RequestInterceptor` protocol works correctly but is only used at the collector/callback level in Journey and DaVinci. Updated doc comments clarifying its scope

### DIFF
--- a/Orchestrate/Orchestrate/RequestInterceptor.swift
+++ b/Orchestrate/Orchestrate/RequestInterceptor.swift
@@ -12,8 +12,23 @@
 import Foundation
 import PingNetwork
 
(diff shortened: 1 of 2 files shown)


george-bafaloukas-forgerock (approved):
LGTM

PR #139, opened 2026-03-26
SDKS-4850_DeviceBinding_name_id

george-bafaloukas-forgerock:
Aligning the DeviceBinding name and allowing to set a custom identifier as well

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-SDKS-4850) "DeviceBinding_name_id"

# Description

# Checklist:

### DIFF
--- a/Binding/Binding/Config/DeviceBindingConfig.swift
+++ b/Binding/Binding/Config/DeviceBindingConfig.swift
@@ -2,13 +2,14 @@
 //  DeviceBindingConfig.swift
 //  PingBinding
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
 //
 
 import Foundation
+import PingDeviceId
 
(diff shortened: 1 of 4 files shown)


Binding/Binding/Config/DeviceBindingConfig.swift
```diff
@@ -20,18 +21,18 @@ public class DeviceBindingConfig {
     #if canImport(UIKit)
     /// Helper private variable for deviceName
     private var _deviceName: String?
-    /// The name of the device.
-    /// The default is the current device name.
+    /// The human-readable name for the device that will be displayed in user interfaces.
+    /// Defaults to the device model (e.g. "iPhone"), matching Android's Build.MODEL default.
```
spetrov, line 25: The "matching Android's Build.MODEL default." is a bit confusing at first read. 
I guess this means the "Android SDK's default behaviour", right?

spetrov (approved):
LGTM! 👍🏻

Binding/Binding/Config/DeviceBindingConfig.swift
george-bafaloukas-forgerock: That's the property used in Android `Build.MODEL`

vahancouver (approved):
LGTM

PR #141, opened 2026-03-30
SDKS-4682"Polling Wait"

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4682) "Polling Wait"

# Description

### Summary of Changes
**New Collectors**:

**PollingCollector**: Supports two modes: Simple (local timer/retry) and Challenge (active polling of a server endpoint). Includes logic to persist retry counts across "rewind" events using SharedContext.

**QRCodeCollector**: A display-only collector that decodes Base64 data URIs into Data for image rendering.

**Rewind Logic**: Updated NodeTransformModule to handle rewindStateToLastRenderedUI and rewindStateToSpecificRenderedUI events. It now re-creates fresh Connector instances, ensuring collectors like PollingCollector reset their internal state appropriately.

**SDK Integration**: 
Registered the new collectors in DaVinci.swift and updated CollectorFactory to inject DaVinci instances into collectors implementing DaVinciAware.

**Sample App**: 
Added PollingCollectorView (with auto-submission logic) and QRCodeCollectorView to demonstrate best practices for these components in SwiftUI.

**Infrastructure**: 
Updated project files, dependencies (added PingDeviceId to PingBinding), and added comprehensive unit tests for the new collectors and rewind scenarios.

# Checklist:


Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,335 @@
...
+import PingOrchestrate
+import PingNetwork
+
+/// Represents the status of a polling operation.
+public enum PollingStatus: Sendable {
+    /// Polling is in progress.
+    /// - Parameters:
+    ///   - retryCount: The current retry count (1-based).
+    ///   - maxRetries: The maximum number of retries configured.
+    case continuing(retryCount: Int, maxRetries: Int)
```
witrisna: May we change it to `continue` instead, the name with align with what we send to server.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,335 @@
...
+        didSet {
+            guard let node = continueNode else { return }
+
+            // Restore retry count persisted by a previous polling cycle after a rewind event.
+            if let remaining = node.context.flowContext
+                .get(key: SharedContext.Keys.pollingRetriesRemaining) as? Int {
+                retriesAllowed = remaining
+            }
+
+            // The server places `pollChallengeStatus` and `challenge` at the root of the
```
witrisna: The POLLING type has those attributes, use the one from the field instead from the Node
```
{
          "type": "POLLING",
          "key": "polling-field",
          "pollInterval": "2000",
          "pollRetries": "20",
          "pollChallengeStatus": true,
          "challenge": "NS3YWOZ_h_MXGoC1j4-IOQLqu3ujXvawPiKBLySe2wg"
}
```

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    // MARK: - Private
+
+    private func pollForChallengeStatus(continuation: AsyncStream<PollingStatus>.Continuation) async {
+        guard
+            let node = continueNode,
+            let links = node.input[Constants._links] as? [String: Any],
+            let next = links[Constants.next] as? [String: Any],
+            let nextHref = next[Constants.href] as? String,
```
witrisna: For Android, I have changed it to use selfHref instead of nextHref.

witrisna (changes requested):

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,335 @@
...
+    /// The node is kept alive by the view layer that is currently displaying it, so the weak
+    /// reference is always valid for the duration of an active polling session.
+    /// When set, restores `retriesAllowed` from FlowContext if a value was persisted by a previous
+    /// polling cycle (i.e. after a rewindStateToLastRenderedUI event creates a fresh collector).
+    public weak var continueNode: ContinueNode? {
+        didSet {
+            guard let node = continueNode else { return }
+
+            // Restore retry count persisted by a previous polling cycle after a rewind event.
+            if let remaining = node.context.flowContext
```
witrisna: instead of using FlowContext to keep track of the `retriesAllowed` state, can we just use instance variable instead? For using flowContext, we may need to consider a flow with multiple continue polling (In different Node but not the same Node)

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: That was as a backup as something was not working for me earlier. I seem to not need that at all, I removed it.

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: `continue` is a keyward in swift so we cannot use it. It's either that or `_ continue` which is not my preferrence

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: We cannot use an instance variable as this will get overriden on every try. In iOS in order to re-render we need to re-create the collector. You are right though the way this was handled if we had multiple connectors in the same flow it will cause an issue. This has been addressed by taking into account the `connectorId` that will be unique.

Davinci/Davinci/module/Transform.swift
```diff
@@ -81,6 +81,27 @@ public class NodeTransformModule {
             return SuccessNode(input: json, session: SessionResponse(json: json))
         }
         
+        // Handle rewindStateToLastRenderedUI and rewindStateToSpecificRenderedUI:
+        // Re-create a fresh Connector from the stored node's input. This intentionally resets
+        // all collector state (including PollingCollector.retriesAllowed) back to its initial
+        // JSON values, giving the user a clean retry. The new ObjectIdentifier on each fresh
+        // collector lets the SwiftUI layer detect the change and restart any running .task.
+        if let eventName = json[Constants.eventName] as? String,
+           (eventName == "rewindStateToLastRenderedUI" || eventName == "rewindStateToSpecificRenderedUI") {
```
vahancouver: can we make these strings to Constants

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,340 @@
...
+    /// The DaVinci workflow instance providing access to the HTTP client.
+    public var davinci: DaVinci?
+
+    // MARK: - Properties
+
+    /// Polling interval in milliseconds between each attempt. Default: `"2000"`.
+    public private(set) var pollInterval: String = "2000"
+
+    /// Maximum number of polling attempts before timing out. Default: `"60"`.
+    public private(set) var pollRetries: String = "60"
```
vahancouver: we have the value 60 and 2000 in multiple places, both as String and Int. Can we extract them to Constnats  as Int then use String(Constants.IntValue) wherever we need the String value

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,340 @@
...
+            let interactionId = node.input[Constants.interactionId] as? String
+        else {
+            value = Constants.pollingValueError
+            continuation.yield(.error(PollingError.missingConfiguration))
+            return
+        }
+
+        // Derive the HTTP client from the workflow stored in the ContinueNode rather than
+        // relying on DaVinciAware injection, which may not fire for closure-registered collectors.
+        let httpClient = node.workflow.config.httpClient!
```
vahancouver: do we have a way not to use force Unwrap here?

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,340 @@
...
+                let response = try await httpClient.request { req in
+                    req.url = pollingUrl
+                    req.setHeader(name: Constants.interactionId, value: interactionId)
+                    req.post(json: [:])
+                }
+
+                // HTTP 400 means the challenge has expired on the server side.
+                // Other non-200 responses are transient — keep polling.
+                guard response.status == 200 else {
+                    if response.status == 400 {
```
vahancouver: In Network Module it's advised to use response.status.isSuccess() and response.status.isClientError() methods instead of comparing to status codes

vahancouver (changes requested):
Added some minor comments.

Also in the UI the retry count goes over max retries, is it supposed to do that? Or is it because it's a number challenge? 

Image

SampleApps/PingExample/PingExample/DavinciViewModel.swift
```diff
@@ -32,7 +32,14 @@ public let davinci = DaVinci.createDaVinci { config in
         oidcValue.scopes = Set<String>(currentConfig?.scopes ?? [])
         oidcValue.redirectUri = currentConfig?.redirectUri ?? ""
         oidcValue.discoveryEndpoint = currentConfig?.discoveryEndpoint ?? ""
-        oidcValue.acrValues = "" //update with actual ACR values if needed or remove
+        oidcValue.acrValues = "28f69709ab22084e4a29c8b18cf574fe " //update with actual ACR values if needed or remove
```
witrisna: May not push local test configuration.

SampleApps/PingExample/PingExample/DavinciViewModel.swift
```diff
@@ -170,6 +177,9 @@ class DavinciViewModel: ObservableObject {
 /// A model class that represents the state of the current and previous nodes in the DaVinci flow.
 class DavinciState {
     var node: Node? = nil
+    /// Unique identifier for this state instance, used to force SwiftUI view recreation
+    /// when the same ContinueNode is returned (e.g. after a rewindStateToLastRenderedUI event).
+    let id = UUID()
```
witrisna: Do u still need this?

SampleApps/PingExample/PingExample/DavinciViewModel.swift
george-bafaloukas-forgerock: How did that get up there :D 
Will remove

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: yeah nice, missed that

george-bafaloukas-forgerock:
> Added some minor comments.
> 
> Also in the UI the retry count goes over max retries, is it supposed to do that? Or is it because it's a number challenge?
> 
> Image

Yeah that's to be expected as behavior based on how DV works at the moment

vahancouver (approved):
Changes LGTM

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+
+    public required init(with json: [String: Any]) {
+        super.init(with: json)
+        // Accept both String ("2000") and numeric (2000) JSON representations so the collector
+        // is robust regardless of whether the server quotes these values.
+        pollInterval = Self.jsonString(json, key: Constants.pollInterval) ?? String(Constants.defaultPollInterval)
+        pollRetries  = Self.jsonString(json, key: Constants.pollRetries)  ?? String(Constants.defaultPollRetries)
+        pollChallengeStatus = json[Constants.pollChallengeStatus] as? Bool ?? false
+        challenge = json[Constants.challenge] as? String ?? ""
+        retriesAllowed = Int(pollRetries) ?? Constants.defaultPollRetries
```
witrisna: Will this reset the stored value?

witrisna (approved):

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: Yes in each init of the collector. Also the "stored value" is linked with the id of the collector so it will be unique in the flow in case of many such collectors

Davinci/Davinci/collector/PasswordCollector.swift
```diff
@@ -18,7 +18,8 @@ import PingDavinciPlugin
 /// It is used to collect password data.
 public class PasswordCollector: ValidatedCollector, ContinueNodeAware, Closeable, @unchecked Sendable {
     /// The continue node for the DaVinci flow.
-    public var continueNode: ContinueNode?
```
rodrigoareis, line 21: Update copyright end-year

Davinci/DavinciTests/mock/MockResponse.swift
```diff
@@ -508,4 +508,26 @@ struct MockResponse {
     """.data(using: .utf8)!
     }
     
+    static var rewindStateToLastRenderedUIResponse: Data {
```
rodrigoareis, line 511: Update copyright end-year

DavinciPlugin/DavinciPlugin/Constants.swift
```diff
@@ -30,6 +30,7 @@ public enum Constants {
     public static let category = "category"
     public static let next = "next"
     public static let href = "href"
+    public static let _self = "self"
```
rodrigoareis, line 33: Update copyright end-year

SampleApps/PingExample/PingExample/ContinueNodeView.swift
```diff
@@ -71,6 +71,15 @@ struct ContinueNodeView: View {
                     DeviceAuthenticationView(field: deviceAuthenticationCollector, onNext: onNext)
                 case let phoneNumberCollector as PhoneNumberCollector:
                     PhoneNumberView(field: phoneNumberCollector, onNodeUpdated: onNodeUpdated)
+                case let pollingCollector as PollingCollector:
```
rodrigoareis, line 74: Update copyright end-year

SampleApps/PingExample/PingExample/DavinciViewModel.swift
```diff
@@ -32,7 +32,14 @@ public let davinci = DaVinci.createDaVinci { config in
         oidcValue.scopes = Set<String>(currentConfig?.scopes ?? [])
         oidcValue.redirectUri = currentConfig?.redirectUri ?? ""
         oidcValue.discoveryEndpoint = currentConfig?.discoveryEndpoint ?? ""
-        oidcValue.acrValues = "" //update with actual ACR values if needed or remove
```
rodrigoareis: Update copyright end-year

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+/// completes, expires, times out, or an error occurs.
+///
+/// ## Value Assignment
+/// The `value` property is updated automatically before each terminal status is yielded:
+/// - `.complete` → the server status string (e.g. `"approved"` or `"continue"`)
+/// - `.timedOut` → `"timedOut"`
+/// - `.expired` → `"expired"`
+/// - `.error` → `"error"`
+/// - `.continuing` → `"continue"` 
+public class PollingCollector: SingleValueCollector, Submittable, ContinueNodeAware, DaVinciAware, Closeable, @unchecked Sendable {
```
rodrigoareis, line 65: The class declares `DaVinciAware` conformance and has `public var davinci: DaVinci?`, but `davinci` is never read anywhere.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+}
+
+/// Errors specific to `PollingCollector` operations.
+public enum PollingError: Error, LocalizedError, Sendable {
+    /// Required configuration (`nextHref` or `interactionId`) is missing from the continue node.
+    case missingConfiguration
+    /// The `pollInterval` value is invalid or non-positive.
+    case invalidInterval
+    /// The polling response body could not be parsed as JSON.
+    case invalidResponse
```
rodrigoareis, line 346: The `.invalidResponse` case is never thrown or used anywhere. We should remove the unused case or use it.

DavinciPlugin/DavinciPlugin/Constants.swift
```diff
@@ -90,4 +91,25 @@ public enum Constants {
...
+    public static let pollInterval = "pollInterval"
+    public static let pollRetries = "pollRetries"
+    public static let pollChallengeStatus = "pollChallengeStatus"
+    public static let challenge = "challenge"
+    public static let isChallengeComplete = "isChallengeComplete"
+    public static let rewindStateToLastRenderedUI = "rewindStateToLastRenderedUI"
+    public static let rewindStateToSpecificRenderedUI = "rewindStateToSpecificRenderedUI"
+    public static let interactionId = "interactionId"
+    public static let fallbackText = "fallbackText"
+    public static let base64Image = "base64Image"
```
rodrigoareis: `base64Image` is declared but never used anywhere in the codebase.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+/// `{baseUrl}/davinci/user/credentials/challenge/{challenge}/status` until the challenge
+/// completes, expires, times out, or an error occurs.
+///
+/// ## Value Assignment
+/// The `value` property is updated automatically before each terminal status is yielded:
+/// - `.complete` → the server status string (e.g. `"approved"` or `"continue"`)
+/// - `.timedOut` → `"timedOut"`
+/// - `.expired` → `"expired"`
+/// - `.error` → `"error"`
+/// - `.continuing` → `"continue"`
```
rodrigoareis:
```suggestion
/// - `.continue` → `"continue"` 
```

DavinciPlugin/DavinciPlugin/Constants.swift
```diff
@@ -90,4 +91,25 @@ public enum Constants {
     public static let phoneNumber = "phoneNumber"
     public static let defaultCountryCode = "defaultCountryCode"
     public static let validatePhoneNumber = "validatePhoneNumber"
+    public static let POLLING = "POLLING"
```
rodrigoareis, line 96: As the constants in this file are public, I would recommend at minimum, group them. For the new ones we could do something like:

```
// MARK: - Polling & QR Code 
```

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+        // Derive the HTTP client from the workflow stored in the ContinueNode rather than
+        // relying on DaVinciAware injection, which may not fire for closure-registered collectors.
+        guard let httpClient = node.workflow.config.httpClient else {
+            value = Constants.pollingValueError
+            continuation.yield(.error(PollingError.missingConfiguration))
+            return
+        }
+
+        let baseUrl = selfHref.components(separatedBy: Constants.davinciConnectionsPathSegment).first ?? selfHref
+        let pollingUrl = "\(baseUrl)\(Constants.challengeStatusPathPrefix)\(challenge)\(Constants.challengeStatusPathSuffix)"
```
rodrigoareis: The `challenge` is included directly into the URL without any URL-encoding or validation. If the server ever returns a challenge value with special characters (e.g., /, ?, #, %), this could produce a malformed or unintended URL. Consider using `URLComponents`.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+                let response = try await httpClient.request { req in
+                    req.url = pollingUrl
+                    req.setHeader(name: Constants.interactionId, value: interactionId)
+                    req.post(json: [:])
+                }
+
+                // HTTP 400 means the challenge has expired on the server side.
+                // Other non-200 responses are transient — keep polling.
+                guard response.status.isSuccess() else {
+                    if response.status.isClientError() {
```
rodrigoareis, line 245: I believe this logic matches all `4xx` status codes (401, 403, 404, 409, etc.). This means a `401 Unauthorized` would also be reported as `.expired` rather than `.error`. Consider checking specifically for status code 400 instead.

SampleApps/PingExample/PingExample/DavinciViewModel.swift
```diff
@@ -32,7 +32,14 @@ public let davinci = DaVinci.createDaVinci { config in
...
+        oidcValue.acrValues = "ACR_VALUE" //update with actual ACR values if needed or remove
```
rodrigoareis: I think this should remain "" or capture from config to avoid confusion.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+/// - `.continuing` → `"continue"` 
+public class PollingCollector: SingleValueCollector, Submittable, ContinueNodeAware, DaVinciAware, Closeable, @unchecked Sendable {
+
+    // MARK: - ContinueNodeAware
+
+    /// The continue node providing configuration context (links, interactionId) for challenge polling.
+    /// Declared `weak` to break the retain cycle: `ContinueNode → PollingCollector → ContinueNode`.
+    /// The node is kept alive by the view layer that is currently displaying it, so the weak
+    /// reference is always valid for the duration of an active polling session.
+    public weak var continueNode: ContinueNode? {
```
rodrigoareis, line 77: The class is marked `@unchecked Sendable` but has some mutable properties, such as `continueNode`.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    // MARK: - Polling
+
+    /// Starts polling and returns an `AsyncStream` of `PollingStatus` updates.
+    ///
+    /// The stream finishes when polling reaches a terminal state
+    /// (`.complete`, `.timedOut`, `.expired`, or `.error`).
+    /// The `value` property is updated automatically before each status is yielded.
+    public func poll() -> AsyncStream<PollingStatus> {
```
rodrigoareis, line 155: This method creates an unstructured `Task` that reads and writes the class properties concurrently with potential SwiftUI main-actor access. I think we may have a latent data-race risk under strict concurrency checking.

Davinci/Davinci/collector/PollingCollector.swift
```diff
@@ -0,0 +1,349 @@
...
+
+    // MARK: - DaVinciAware
+
+    /// The DaVinci workflow instance providing access to the HTTP client.
+    public var davinci: DaVinci?
+
+    // MARK: - Properties
+
+    /// Polling interval in milliseconds between each attempt. Default: `"2000"`.
+    public private(set) var pollInterval: String = String(Constants.defaultPollInterval)
```
rodrigoareis: `pollInterval` and `pollRetries` are stored as `String` but repeatedly parsed with `Int(pollRetries)` and `Double(pollInterval)` in multiple methods. I think storing them as `Int` / `Double` after parsing once during `init` would be cleaner and avoid repeated parsing.

Davinci/Davinci/collector/QRCodeCollector.swift
```diff
@@ -0,0 +1,68 @@
...
+/// `"data:image/png;base64,iVBORw0KGgo..."`
+///
+/// The `imageData` property exposes the decoded `Data` after stripping the data URI prefix,
+/// allowing the UI layer to render it as a `UIImage` or `SwiftUI.Image`.
+///
+/// The `fallbackText` property carries an alternative string (e.g. a manual entry code or
+/// a URL) to display if the QR code cannot be rendered or scanned.
+///
+/// This collector does not participate in form submission — `payload()` always returns `nil`.
+public class QRCodeCollector: Collector, @unchecked Sendable {
```
rodrigoareis, line 26: It directly conforms to Collector, but if more display-only collectors are added in the future, a `DisplayCollector` base class could reduce boilerplate.

SampleApps/PingExample/PingExample/Collectors/PollingCollectorView.swift
```diff
@@ -0,0 +1,130 @@
...
+///   `FlowContext` so the counter continues from where it left off.
+///
+/// - `.complete(status: other)` (challenge polling succeeded) — `onNext` submits the server
+///   status and the flow advances to the next step.
+///
+/// - `.timedOut` / `.expired` — the respective terminal value is submitted so the server can
+///   decide the next step.
+///
+/// - `.error` — the error is surfaced to the user. No automatic submission is made.
+struct PollingCollectorView: View {
```
rodrigoareis: Nitpick comment: The existing collector views use short names without `"Collector"` (e.g., TextView, PasswordView, LabelView), but the new views use the full name (`PollingCollectorView`, `QRCodeCollectorView`). Other recent views also use the full name. I would suggest we align it, even if this just a sample app.

DavinciPlugin/DavinciPlugin/Constants.swift
rodrigoareis:
```suggestion
//  PingDavinciPlugin
```

Davinci/DavinciTests/QRCodeCollectorTests.swift
```diff
@@ -0,0 +1,134 @@
...
+//
+
+import XCTest
+@testable import PingDavinci
+
+class QRCodeCollectorTests: XCTestCase {
+
+    // MARK: - Initialization
+
+    func testInitializesWithDataURIContent() {
```
rodrigoareis, line 18: No test for malformed base64 data

Davinci/DavinciTests/DaVinciTests.swift
```diff
@@ -460,4 +460,94 @@ final class DaVinciTests: DaVinciBaseTests, @unchecked Sendable {
         XCTAssertEqual(collector11.options.count, 2)
         XCTAssertEqual(collector11.value, ["default-checkbox"])
     }
+    
+    func testDaVinciRewindStateToLastRenderedUIReturnsPreviousContinueNode() async throws {
+        MockURLProtocol.requestHandler = { request in
+            switch request.url!.path {
+            case MockAPIEndpoint.discovery.url.path:
+                return (HTTPURLResponse(url: MockAPIEndpoint.discovery.url, statusCode: 200, httpVersion: nil, headerFields: MockResponse.headers)!, MockResponse.openIdConfigurationResponse)
```
rodrigoareis: force-unwrap, while acceptable in tests, a guard with XCTFail is safer

Davinci/DavinciTests/DaVinciTests.swift
```diff
@@ -460,4 +460,94 @@ final class DaVinciTests: DaVinciBaseTests, @unchecked Sendable {
...
+            case MockAPIEndpoint.authorization.url.path:
+                return (HTTPURLResponse(url: MockAPIEndpoint.authorization.url, statusCode: 200, httpVersion: nil, headerFields: MockResponse.authorizeResponseHeaders)!, MockResponse.authorizeResponse)
+            case MockAPIEndpoint.customHTMLTemplate.url.path:
+                // `start()` gets its ContinueNode from /authorize; `next()` is the first (and only)
+                // call to this URL and must return the rewind response.
+                return (HTTPURLResponse(url: MockAPIEndpoint.customHTMLTemplate.url, statusCode: 200, httpVersion: nil, headerFields: MockResponse.customHTMLTemplateHeaders)!, MockResponse.rewindStateToLastRenderedUIResponse)
+            default:
+                return (HTTPURLResponse(url: MockAPIEndpoint.discovery.url, statusCode: 500, httpVersion: nil, headerFields: nil)!, Data())
```
rodrigoareis: force-unwrap

Package.swift
```diff
@@ -298,6 +298,7 @@ let package = Package(
             name: "PingBinding",
             dependencies: [
                 "PingCommons",
+                "PingDeviceId",
```
rodrigoareis, line 312: Is this intentional?

PingBinding.podspec
```diff
@@ -22,5 +22,6 @@ Pod::Spec.new do |s|
   }
   
   s.ios.dependency 'PingCommons', '~> 1.3.1'
+  s.ios.dependency 'PingDeviceId', '~> 1.3.1'
```
rodrigoareis: Is this intentional?

rodrigoareis (changes requested):
Overall changes looks good to me. Left some comments do be addressed.

Package.swift
george-bafaloukas-forgerock: yes, this was missing

PingBinding.podspec
george-bafaloukas-forgerock: yes, this was missing. As part of the release those will be cleaned and aligned

Davinci/Davinci/collector/QRCodeCollector.swift
george-bafaloukas-forgerock: This is not a bad suggestion but we are now aligned with Android. I think we should keep that in mind and implement the new protocol when and if needed in both platforms

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: Any non 200 responses are considered expired. So this check should cover the use case. 
This is according to the Android implementation `// Handle non-200 HTTP responses as expired`

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: Good catch, fixed

Davinci/DavinciTests/DaVinciTests.swift
george-bafaloukas-forgerock: This and the other force-unwrap is touching on the nitpicking for the tests as we have this pattern in may places. In any case, these have now been improved across the file

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: I'll keep it to have parity with Android. I am also adding use of the logger as set in the Davinci Config, so this is now used. The main difference with Android is that we get the httpClient from the workFlow config

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: The Sendable conformance is already baked into the base class. Removing it from the PollingCollector declaration just removes a redundant annotation — the compiler still considers it Sendable through inheritance, and @unchecked still suppresses all diagnostics because that's what the superclass declared.

To actually get the compiler to flag data races, you'd need to remove @unchecked Sendable from the entire FieldCollector → SingleValueCollector hierarchy — which would break every collector in the SDK, since they all have mutable properties (value, key, etc.) and none use internal synchronization.

Since we have this pattern across the collectors I think we are safe for now. A general big refactoring on this would have to implement locks and conform to sendable across all collectors which will be a very big task.

**In the current code**:

pollForChallengeStatus reads it once at the top with guard let node = continueNode and uses the strong local node for the entire method — safe regardless of when ARC zeros the weak ref.

pollSimple reads it via optional chaining (continueNode?.input, continueNode?.context.flowContext.set(...)) — each access independently loads a strong reference or gets nil. No crash risk, though theoretically the two optional-chained accesses could see different states (first succeeds, second gets nil if the node was deallocated between them). In practice this doesn't matter because if the node is gone mid-poll, the flow is already torn down.

The didSet observer writes retriesAllowed, but this runs during node construction before poll() starts — no overlap.

So there's no safety issue with continueNode itself. The weak reference semantics are handled by the Swift runtime, and the usage pattern (set once before polling, read during polling) avoids any logical race.

Davinci/Davinci/collector/PollingCollector.swift
george-bafaloukas-forgerock: That's an interesting one. I played around with a few options but here is the analysis:
<h2 style="line-height: normal; font-size: 16.003px; font-weight: 600; margin: 1.5em 0px 0.875em; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; color: rgb(191, 191, 191); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The Issue</h2><p style="margin: 0px 0px 16px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">poll()</code><span> </span>creates an unstructured<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">Task</code><span> </span>that reads/writes mutable properties (<code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">value</code>,<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">retriesAllowed</code>) on the cooperative thread pool, while the SwiftUI consumer (<code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">PollingView</code>) accesses the same object from<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">@MainActor</code><span> </span>context.</p><p style="margin: 0px 0px 16px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong style="font-weight: 600;">Properties at risk:</strong></p><div class="monaco-scrollable-element mac rendered-markdown-table-scroll-wrapper" role="presentation" style="width: 477px; margin-bottom: 16px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; position: relative; overflow: hidden;"><div style="overflow: hidden;">
Property | Written by poll Task | Read by MainActor
-- | -- | --
value | Yes — set before every continuation.yield() | Yes — via payload() when onNext triggers form submission
retriesAllowed | Yes — decremented in pollSimple | Yes — read in PollingView.init() to seed the counter
continueNode | Read (weak) | Set via didSet from workflow setup

</div><div role="presentation" aria-hidden="true" class="invisible scrollbar horizontal" style="opacity: 0; pointer-events: none; background: none 0% 0% / auto repeat scroll padding-box border-box rgba(0, 0, 0, 0); position: absolute; width: 477px; height: 10px; left: 0px; bottom: 0px;"><div class="slider" style="background: none 0% 0% / auto repeat scroll padding-box border-box rgba(131, 132, 133, 0.2); position: absolute; top: 0px; left: 0px; height: 10px; transform: translate3d(0px, 0px, 0px); contain: strict; width: 477px;"></div></div><div role="presentation" aria-hidden="true" class="invisible scrollbar vertical" style="opacity: 0; pointer-events: none; background: none 0% 0% / auto repeat scroll padding-box border-box rgba(0, 0, 0, 0); z-index: 14; position: absolute; width: 0px; height: 196px; right: 0px; top: 0px;"><div class="slider" style="background: none 0% 0% / auto repeat scroll padding-box border-box rgba(131, 132, 133, 0.2); position: absolute; top: 0px; left: 0px; width: 10px; transform: translate3d(0px, 0px, 0px); contain: strict; height: 196px;"></div></div></div><h2 style="line-height: normal; font-size: 16.003px; font-weight: 600; margin: 1.5em 0px 0.875em; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; color: rgb(191, 191, 191); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Why It Works Today</h2><ol style="padding-inline-start: 28px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li style="margin: 4px 0px;"><strong style="font-weight: 600;">Sequential usage</strong><span> </span>—<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">poll()</code><span> </span>is called exactly once per collector instance from<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">.task { }</code>, and<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">onNext()</code><span> </span>is called only after<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">for await</code><span> </span>yields, so the AsyncStream acts as a<span> </span><em>de facto</em><span> </span>synchronization barrier between the<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">value</code><span> </span>write and the<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">value</code><span> </span>read.</li><li style="margin: 4px 0px;"><strong style="font-weight: 600;"><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">retriesAllowed</code></strong><span> </span>read in<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">PollingView.init()</code><span> </span>happens<span> </span><em>before</em><span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">.task</code><span> </span>starts polling — no concurrent access.</li><li style="margin: 4px 0px;"><strong style="font-weight: 600;"><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">continueNode</code></strong><span> </span>is set during node construction, before the view is displayed — no overlap with<span> </span><code style="font-family: &quot;SF Mono&quot;, Monaco, Menlo, Courier, monospace; font-size: 10.998px; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 1px 3px; border-radius: 4px; border-color: rgb(140, 140, 140); border-style: none; border-width: 0px; border-image: none 100% / 1 / 0 stretch; white-space: pre-wrap;">poll()</code>.</li></ol>

george-bafaloukas-forgerock: Now the fact is this very rarely would go wrong but in any case I added a lock that protect this from any racing. Adding a lock on this is the most straightforward approach

spetrov (approved):
LGTM! 👍🏻

rodrigoareis (approved):
LGTM

PR #142, opened 2026-03-30
Sdks 4893 device client fixes

vahancouver:
[SDKS-4893](https://pingidentity.atlassian.net/browse/SDKS-4893): [Ping SDK][iOS] Sending value of non-Sendable type risks causing data races

### DIFF
--- a/DeviceClient/DeviceClient/Device.swift
+++ b/DeviceClient/DeviceClient/Device.swift
@@ -13,7 +13,7 @@ import Foundation
 import PingCommons
 
 /// Protocol defining device operations.
-public protocol DeviceRepository<T> {
+public protocol DeviceRepository<T>: Sendable {
     associatedtype T
     
     /// Retrieves a list of devices.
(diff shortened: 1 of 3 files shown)


DeviceClient/DeviceClient/Device.swift
spetrov: Update copyright header

spetrov (approved):
LGTM! 👍🏻

DeviceClient/DeviceClient/Device.swift
george-bafaloukas-forgerock: updated

george-bafaloukas-forgerock (approved):
LGTM, not a major fan of the need to use a `nonisolated(unsafe)` but this might be the only pragmatic way to bridge the gap between the system callbacks and modern Task blocks. Let's monitor this in future releases to see if we can make any improvements in the DeviceClient

george-bafaloukas-forgerock:
@forgerock-chris This will definitely need to be called out in the docs. Make a note please for when you start working on the device client and Push

PR #143, opened 2026-04-02
Add XCFrameworks support

vahancouver:
[SDKS-2889](https://pingidentity.atlassian.net/browse/SDKS-2889)
[SPIKE] [iOS] Create a Helper Docs for XCFrameworks

This addresses swift language issues that occur during XCFramework generations.

* Make transitive imports explicit 
   - `import Combine` wherever `ObservableObject` is used, 
   -  import the module that has typealiased type - `PingNetwork`, `PingOrchetsrate`, `PingDeviceId`
* Update the scheme names to match the Module names (mainly to be able to use a script)
* Mark 3rd party library imports as internal - This ensures that the imported module is only accessible within the current module.  It avoids "leaking" dependencies to consumers of the XCFramework, which helps prevent issues like "No such module" errors when clients try to build against our binary.

To test the XCFramework generation:
- checkout this branch,
- restart Xcode (this is needed for Xcode to pickup the updated scheme names
- download and save the [build-xcframeworks.sh](https://github.com/user-attachments/files/26578616/build-xcframeworks.sh) file in the root folder and run `./build-xcframeworks.sh`

more info: https://pingidentity.atlassian.net/wiki/spaces/SDKS/pages/2731606076/SDKS-2889+XCFramework+Distribution+Guide


rodrigoareis (approved):
Changes looks good to me

george-bafaloukas-forgerock (approved):
LGTM

spetrov (approved):
LGTM! 👍🏻

PR #144, opened 2026-04-08
SDKS-4773 AuthMigration Module 

rodrigoareis:
# JIRA Ticket

[SDKS-4773](https://pingidentity.atlassian.net/browse/SDKS-4773) "AuthMigration Module "

# Description

### Summary

Introduces a new `PingAuthMigration` module that migrates OATH (TOTP/HOTP) and Push credentials from the legacy `FRAuthenticator` SDK Keychain format to the modern `PingOath` / `PingPush` storage format.

### Changes

**New Module: `PingAuthMigration`**
- `AuthMigration.swift` – Public API with `start(_:)`, `migrateIfNeeded()`, and `isMigrationNeeded()` entry points
- `AuthMigrationConfig.swift` – DSL-style configuration (access group, logger, cleanup flag, custom storage)
- `AuthMigrationError.swift` – Typed error cases for all failure scenarios
- `AuthMigrationStep+Extensions.swift` – Human-readable step descriptions
- `LegacyKeychainReader.swift` – Reads legacy FRAuthenticator Keychain items, including Secure Enclave–encrypted data
- `LegacyArchiveModels.swift` – NSKeyedUnarchiver-based legacy data models
- `LegacyDataConverter.swift` – Converts legacy mechanisms to `OathCredential` / `PushCredential`
- `LegacySecuredKey.swift` – Handles Secure Enclave key operations for decrypting legacy data

**`PingCommons` additions**
- `MigrationProgress.swift` and `MigrationStep.swift` – Shared `AsyncStream<MigrationProgress>` protocol used to report migration progress to callers

**Sample App**
- `AuthMigrationView.swift` / `AuthMigrationViewModel.swift` – Example UI demonstrating progress-tracked migration in `PingExample`

### Testing

Full XCTest suite added under `AuthMigration/AuthMigrationTests/`, covering:
- OATH and Push credential conversion
- Legacy Keychain reading (including encrypted paths)
- Config, error types, and step extensions
- Mock OATH/Push storage and Keychain helpers

# Checklist:

> [!NOTE]
> **For reviewers:** The diff is large because it includes commits for the 2.0 release. Please focus on the AuthMigration module.


AuthMigration/AuthMigration/Internal/LegacySecuredKey.swift
```diff
@@ -0,0 +1,120 @@
...
+    /// - Throws: ``AuthMigrationError/failedToDecryptLegacyData(_:)`` if both algorithms fail.
+    func decrypt(_ data: Data) throws -> Data {
+        // Try current algorithm first
+        var error: Unmanaged<CFError>?
+        if let decrypted = SecKeyCreateDecryptedData(privateKey, Self.currentAlgorithm, data as CFData, &error) {
+            return decrypted as Data
+        }
+
+        // Fallback to legacy algorithm
+        error = nil
```
george-bafaloukas-forgerock: Potential memory leak here:
`SecKeyCreateDecryptedData` populates the error pointer with an Unmanaged<CFError> containing a +1 retain count. By setting error = nil without releasing it, you strand the object in memory.

Suggested fix:
```
if let cfError = error {
    cfError.release()
    error = nil
}
```

AuthMigration/AuthMigration/AuthMigration.swift
```diff
@@ -0,0 +1,386 @@
...
+            continuation.yield(.error(step: .importLegacyData, underlyingError: AuthMigrationError.noLegacyDataFound))
+            continuation.finish()
+            return false
+        }
+
+        logger?.i("Found \(accounts.count) accounts and \(mechanisms.count) mechanisms")
+
+        // Build account lookup: "issuer-accountName" → LegacyAccountArchive
+        let accountLookup = Dictionary(
+            accounts.map { ("\($0.issuer)-\($0.accountName)", $0) },
```
george-bafaloukas-forgerock, line 240: Bit of an edge case but we can have a potential Dictionary Key Collision. If an issuer or account name contains a hyphen, we risk a collision. For example, Issuer "My", Account "App-User" and Issuer "My-App", Account "User" will both resolve to "My-App-User".
Maybe use a safer delimiter (like "::" or ||) or create a quick Hashable struct to act as the compound key.

ExternalIdP/ExternalIdP.xcodeproj/project.pbxproj
```diff
@@ -359,7 +359,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MARKETING_VERSION = "2.0.0-alpha";
+				MARKETING_VERSION = "2.0.0";
```
george-bafaloukas-forgerock: Not caused by your change but I see we have a typo on the bundle identifier in the `com.pingidentity.ExtrernalIdP` we might want to fix this before the 2.0 release

AuthMigration/AuthMigration/Internal/LegacyKeychainReader.swift
```diff
@@ -0,0 +1,275 @@
...
+            unarchiver = try NSKeyedUnarchiver(forReadingFrom: decrypted)
+        } catch {
+            throw AuthMigrationError.failedToDeserializeLegacyData(
+                "Failed to create unarchiver for Account: \(error.localizedDescription)"
+            )
+        }
+
+        unarchiver.requiresSecureCoding = false
+        unarchiver.decodingFailurePolicy = .setErrorAndReturn
+        unarchiver.setClass(LegacyAccountArchive.self, forClassName: "FRAuthenticator.Account")
```
george-bafaloukas-forgerock: Can we move the `FRAuthenticator.Account` in Constants?

AuthMigration/AuthMigration/AuthMigration.swift
```diff
@@ -0,0 +1,386 @@
...
+/// ```
+///
+/// - SeeAlso: ``AuthMigrationConfig``
+/// - SeeAlso: ``AuthMigrationError``
+/// - SeeAlso: `MigrationProgress`
+/// - SeeAlso: `MigrationStep`
+public final class AuthMigration: @unchecked Sendable {
+
+    /// Private actor to ensure mutual exclusion of migration runs.
+    private static let stateManager = MigrationStateManager()
```
george-bafaloukas-forgerock, line 87: We might want to add some tests to ensure that the thread safety is achieved for the MigrationStateManager

Commons/Commons/Migration/MigrationStep.swift
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Steps are passed as associated values in ``MigrationProgress`` events:
+///
+/// ```swift
+/// case .inProgress(let step, let current, let total):
+///     print("Step \(current)/\(total): \(step.description)")
+/// ```
+///
+/// - SeeAlso: ``MigrationProgress``
+public struct MigrationStep: Sendable, CustomStringConvertible {
```
george-bafaloukas-forgerock: Opt in for `Equatable` and `Hashable` protocols as well. This will improve the UI state checks

Commons/Commons/Migration/MigrationStep.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// A human-readable description of the step.
+    ///
+    /// Used in progress reporting and log messages. Should be a short, descriptive phrase
+    /// such as `"Import legacy data"` or `"Migrate credentials"`.
+    public let description: String
```
george-bafaloukas-forgerock, line 56: Maybe it would be better if we add an `id` or `unique` name on the MigrationStep. That way we can separate the need for the `description` to be used for checks in the UI if needed. This will help the `description` also be a display only field that can be translated.

george-bafaloukas-forgerock: In this case as well we can conform to `Identifiable`

george-bafaloukas-forgerock (changes requested):
Great set of changes! I added a comment for a memory leak that needs to be fixed. There are some other suggestions as well for improvement on the migration steps

SampleApps/PingExample/PingExample/ContentView.swift
```diff
@@ -236,6 +240,8 @@ struct ContentView: View {
                     StorageView(menuItem: item)
                 case .bindingKeys:
                     BindingKeysView()
+                case .migration:
+                    AuthMigrationView()
```
vahancouver, line 244: "Migrate legacy SDK data" implies everything but the AuthMigrationView() covers only Auth stuff. should we make it generic and cover Device Binding as well? Or for now specify in the name that this is Auth Migration only.

AuthMigration/AuthMigration.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,582 @@
...
+				SDKROOT = iphoneos;
+				TARGETED_DEVICE_FAMILY = "1,2";
+				TEST_HOST = "$(BUILT_PRODUCTS_DIR)/PingTestHost.app/PingTestHost";
+			};
+			name = Debug;
+		};
+		726A166B7F28BDD475398537 /* Release */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				BUNDLE_LOADER = "$(TEST_HOST)";
```
vahancouver: This needs to be removed... tests are failing to build because of this

AuthMigration/AuthMigration.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,582 @@
...
+				SWIFT_OPTIMIZATION_LEVEL = "-O";
+				SWIFT_UPCOMING_FEATURE_MEMBER_IMPORT_VISIBILITY = YES;
+				SWIFT_VERSION = 6.0;
+			};
+			name = Release;
+		};
+		5D1AB5BFF1A3B63F8E5682D6 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				BUNDLE_LOADER = "$(TEST_HOST)";
```
vahancouver: This needs to be removed... tests are failing to build because of this

vahancouver (changes requested):
Looks great. Small comment to make the test build and run.

Agree with George's comments.

AuthMigration/AuthMigration/Internal/LegacySecuredKey.swift
rodrigoareis: nice catch! The suggested fix `(cfError.release())` would work, but `takeRetainedValue()` is more idiomatic Swift — it transfers ownership to ARC.

AuthMigration/AuthMigration/AuthMigration.swift
rodrigoareis: Agree, this is a pre-existing limitation of the legacy SDK's data model. But here it's intentionally matching the legacy SDK's key format. We must use the same delimiter to correctly join mechanisms to their accounts from the legacy storage. Using a different delimiter (like :: or ||) would actually break the lookup.

ExternalIdP/ExternalIdP.xcodeproj/project.pbxproj
rodrigoareis: @spetrov see this comment.

AuthMigration/AuthMigration/Internal/LegacyKeychainReader.swift
rodrigoareis: I think move it to `Constants` file makes sense when values are shared across multiple files. In here, the Strings are used nowhere else. So I am adding them as `private static let` constants at the top of file alongside the existing service constants.

Commons/Commons/Migration/MigrationStep.swift
rodrigoareis: It's a reasonable improvement,  but I believe it leans toward over-engineering for the current usage. The `id` would just duplicate what the constant name already provides. `MigrationStep` instances are defined as `static let` constants (e.g., `.importLegacyData`). In practice, consumers compare against these constants, not against `description` strings. I'll apply the suggestion anyway.

AuthMigration/AuthMigration/AuthMigration.swift
rodrigoareis: I may be getting it wrong, but my understanding is that thread safety is inherently guaranteed by the Swift actor model — `MigrationStateManager` is an actor, so all its methods are serialized by the runtime. There's nothing to unit test about the concurrency guarantees themselves, right?

That said, we do have behavioral tests for the state machine logic:

`testStart_onlyRunsOnce()` — verifies the `hasRun` guard prevents re-execution
`testResetMigrationState_allowsRerun()` — verifies reset clears the guard
`testStart_secondRunIsNoOp_afterCleanup()` — verifies idempotency after cleanup

One gap I did find: 
we're missing a test for `retry after failure` (where `hasRun` stays false on failure, allowing a subsequent call to execute). I'll add that.

george-bafaloukas-forgerock (approved):
LGTM

vahancouver (approved):
Changes look good

spetrov (approved):
LGTM!

ExternalIdP/ExternalIdP.xcodeproj/project.pbxproj
spetrov: Yes, fixed!

PR #147, opened 2026-04-15
Applying changes to support Binding with fallback when the biometrics…

george-bafaloukas-forgerock:
Applying changes to support Binding with fallback when the biometrics are not set on the device. Adding more tests and code comments explaining the functionality

# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4935) "[iOS][Unified]Investigate DeviceBinding BiometricWithFallback not working when Passcode is set and Biometrics is disabled"

# Description

**Biometric & Passcode Logic Improvements**
Dynamic Access Control: In BiometricDeviceCredentialAuthenticator, the code now checks if biometrics are available before creating keys.

If biometrics are enrolled: Uses [.biometryAny, .or, .devicePasscode].

If no biometrics: Uses only .devicePasscode.

Fix for Creation Failures: By switching from .biometryCurrentSet to .biometryAny, the SDK prevents the Secure Enclave from rejecting key creation on devices where biometrics aren't configured yet.

**2. Security & State Tracking**
Biometric Domain State: Added biometricDomainState (of type Data?) to the UserKey model. This captures a "snapshot" of the device's biometric enrollment at the time of binding.

Tamper Detection: During the sign operation, the authenticator compares the current biometric state with the stored state. If they differ (e.g., a new fingerprint was added), the key is deleted and the user is forced to re-bind for security.

Backward Compatibility: The UserKey initializer now defaults the state to nil, ensuring existing keys created on older versions of the SDK continue to work without triggering the state check.

**3. Project & Housekeeping**
New Tests: Added BiometricDeviceCredentialAuthenticatorTests.swift to the project and build phases.

Documentation: Significantly improved the header documentation in the authenticator class to explain the security model and hardware-vs-software invalidation.

# Checklist:


vahancouver (approved):
Looks good

witrisna (approved):

PR #152, opened 2026-04-24
Add ReadOnlyTextCollector support for Davinci forms

vahancouver:
[SDKS-4928](https://pingidentity.atlassian.net/browse/SDKS-4928)
[iOS] [DaVinci] - Support PingOne Forms Agreements Component
Simulator Screenshot - iPhone Air - 2026-04-24 at 10 15 11

### DIFF
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -94,6 +94,9 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.QR_CODE, closure: { json in
                 return QRCodeCollector(with: json)
             })
+            await CollectorFactory.shared.register(type: Constants.READ_ONLY_TEXT, closure: { json in
+                return AgreementCollector(with: json)
+            })
             if let c: NSObject.Type = NSClassFromString("PingProtect.ProtectCollector") as? NSObject.Type {
                 c.perform(Selector(("registerCollector")))
             }
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -95,7 +95,7 @@ extension DaVinci {
                 return QRCodeCollector(with: json)
             })
             await CollectorFactory.shared.register(type: Constants.READ_ONLY_TEXT, closure: { json in
-                return AgreementCollector(with: json)
+                return ReadOnlyTextCollector(with: json)
             })
             if let c: NSObject.Type = NSClassFromString("PingProtect.ProtectCollector") as? NSObject.Type {
                 c.perform(Selector(("registerCollector")))
--- a/Davinci/Davinci/collector/ReadOnlyTextCollector.swift
+++ b/Davinci/Davinci/collector/ReadOnlyTextCollector.swift
@@ -11,10 +11,8 @@
 import Foundation
 import PingDavinciPlugin
 
-/// Class representing an AGREEMENT type collector.
-///
-/// This class handles the response from a DaVinci form field of type AGREEMENT.
-/// It displays agreement/terms-of-service text and collects the user's acceptance.
+/// This class handles the response from a DaVinci form field of inputType READ_ONLY_TEXT .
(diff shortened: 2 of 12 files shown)


spetrov (approved):
LGTM! 👍🏻

Davinci/Davinci/collector/ReadOnlyTextCollector.swift
```diff
@@ -0,0 +1,90 @@
...
+/// ```
+public class AgreementCollector: Collector, @unchecked Sendable {
+    
+    /// The UUID of the collector.
+    public var id: String {
+        return key
+    }
+    
+    /// The key identifier for this collector, used as the form field key during submission.
+    public private(set) var key: String = ""
```
george-bafaloukas-forgerock, line 41: Do we need to have those as private(set)?
Since these values are only set once during init, consider changing them from `public private(set) var` to `public let`. This ensures complete immutability after initialization. By making them constants, you can likely remove the @unchecked Sendable attribute

Davinci/Davinci/DaVinci.swift
```diff
@@ -94,6 +94,9 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.QR_CODE, closure: { json in
                 return QRCodeCollector(with: json)
             })
+            await CollectorFactory.shared.register(type: Constants.READ_ONLY_TEXT, closure: { json in
```
george-bafaloukas-forgerock, line 97: Just to confirm. Is the type here `AGREEMENT` or we take the `inputType`?
/// {
///   "type": "AGREEMENT",
///   "inputType": "READ_ONLY_TEXT",
///   "key": "agreement",
///   "content": "This is example agreement text...",
///   "titleEnabled": true,
///   "title": "Terms of Service Agreement",
///   "agreement": {
///     "id": "6ff30c9e-cd98-4fe5-85ca-01111ca20702",
///     "useDynamicAgreement": false
///   },
///   "enabled": true
/// }

george-bafaloukas-forgerock (commented):
Overall looks good and solid. Please check the 2 minor comments/questions

Davinci/Davinci/collector/ReadOnlyTextCollector.swift
vahancouver: So I tried that approach, but not sure we are gaining much, here is why:
1. even if all properties are let, it only allows to remove the @unchecked if the class is marked final (which we could do I guess..)
2. we are losing the ability to assign the default values during declaration, as it only allows to initialize once (so it has to be in init
3. in some cases, in init we check if certain criteria is met then assign a value (agreementId, useDynamicAgreement).. now we have to assign values also when the criteria is not met
4. every other Collector we have uses `public private(set) var`, should we start changing now?

Let's discuss

Davinci/Davinci/DaVinci.swift
vahancouver: yes we use "inputType" for this and singlecheckbox

Davinci/Davinci/collector/ReadOnlyTextCollector.swift
george-bafaloukas-forgerock: Sounds good. We might need to review the sendables at some point in all collector/callback cases at a later point

Davinci/Davinci/collector/ReadOnlyTextCollector.swift
```diff
@@ -0,0 +1,90 @@
...
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import Foundation
+import PingDavinciPlugin
+
+/// Class representing an AGREEMENT type collector.
+///
+/// This class handles the response from a DaVinci form field of type AGREEMENT.
+/// It displays agreement/terms-of-service text and collects the user's acceptance.
```
george-bafaloukas-forgerock: Do we actually "collect the user's acceptance." using this? We might want to change this if we don't

george-bafaloukas-forgerock (approved):
LGTM, please have a look on the new comment

PR #153, opened 2026-04-24
SDKS-4235 Implementation of PAR

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4235) "Implementation of PAR"

# Description

## Overview
This PR introduces support for **Pushed Authorization Requests (PAR, RFC 9126)** across the OIDC, DaVinci, and Journey modules. 

When enabled, authorization parameters are securely pushed to the authorization server via a backend POST request rather than being exposed in the frontend redirect URI. This enhances security and prevents URL length issues for complex authorization requests.

## Key Changes
* **OIDC Module Core:** * Added `pushedAuthorizationRequestEndpoint` to `OpenIdConfiguration` parsing.
  * Introduced a `par: Bool` toggle in `OidcClientConfig` (defaults to `false`).
  * Refactored `OidcClient.swift` to support `buildAuthorizeParams`, allowing parameter sharing between standard and PAR flows.
  * Created an async `populateRequest` method that conditionally hits the PAR endpoint, retrieves the `request_uri`, and constructs a secure authorization URL.
* **DaVinci & Journey SDKs:** * Updated modules to seamlessly support the new async `populateRequest` flow.
* **Sample Apps (`PingExample`):** * Added a UI toggle in the Configuration Editor to easily test PAR flows.
* **Testing:** * Added comprehensive unit and integration tests (`DaVinciPARTests.swift`, `OidcClientPARTests.swift`, and additions to `OidcAdditionalTests.swift`) to verify happy paths, missing endpoints, and API failures.

## Fallback Behavior
If PAR is enabled in the configuration but the OpenID discovery document does not contain a `pushed_authorization_request_endpoint`, the SDK will gracefully fall back to the standard authorization flow.

## Reviewer Notes
* The legacy synchronous `populateRequest` remains intact for backward compatibility.

# Checklist:

### DIFF
--- a/Davinci/DavinciTests/DaVinciPARTests.swift
+++ b/Davinci/DavinciTests/DaVinciPARTests.swift
@@ -1,0 +1,280 @@
+//
+//  DaVinciPARTests.swift
+//  DavinciTests
+//
+//  Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import XCTest
+import PingDavinciPlugin
+@testable import PingOrchestrate
+@testable import PingLogger
+@testable import PingOidc
+@testable import PingStorage
+@testable import PingDavinci
+@testable import PingNetwork
+
+final class DaVinciPARTests: XCTestCase, @unchecked Sendable {
+    
+    /// Helper to read httpBodyStream data from URLRequest (httpBody is nil when using URLSession with URLProtocol)
+    private func bodyData(from request: URLRequest) -> Data {
+        if let body = request.httpBody {
+            return body
+        }
+        guard let stream = request.httpBodyStream else {
+            return Data()
+        }
+        stream.open()
+        defer { stream.close() }
+        var data = Data()
+        let bufferSize = 4096
+        let buffer = UnsafeMutablePointer<UInt8>.allocate(capacity: bufferSize)
+        defer { buffer.deallocate() }
+        while stream.hasBytesAvailable {
+            let bytesRead = stream.read(buffer, maxLength: bufferSize)
+            if bytesRead > 0 {
+                data.append(buffer, count: bytesRead)
+            } else {
+                break
+            }
+        }
+        return data
+    }
+    
+    let testClientId = "test"
+    let testScopes = ["openid", "email", "address"]
+    let testRedirectUri = "http://localhost:8080"
+    let testDiscoveryEndpoint = "http://localhost/.well-known/openid-configuration"
+    
+    /// Creates an `HTTPURLResponse` or throws, replacing force-unwrap (`!`) in mock handlers.
+    private func mockResponse(url: URL, statusCode: Int, headers: [String: String]? = nil) throws -> HTTPURLResponse {
+        guard let response = HTTPURLResponse(url: url, statusCode: statusCode, httpVersion: nil, headerFields: headers) else {
+            throw URLError(.badServerResponse)
+        }
+        return response
+    }
+    
+    override func setUp() {
+        super.setUp()
+        MockURLProtocol.startInterceptingRequests()
+        _ = CollectorFactory.shared
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        MockURLProtocol.stopInterceptingRequests()
+    }
+    
+    // MARK: - PAR Happy Path
+    
+    func testDaVinciPARHappyPath() async throws {
+        MockURLProtocol.requestHandler = { [self] request in
+            switch request.url!.path {
+            case MockAPIEndpoint.discovery.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.discovery.url, statusCode: 200, headers: MockResponse.headers), MockResponse.openIdConfigurationWithPARResponse)
+            case MockAPIEndpoint.par.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.par.url, statusCode: 201, headers: MockResponse.headers), MockResponse.parResponse)
+            case MockAPIEndpoint.token.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.token.url, statusCode: 200, headers: MockResponse.headers), MockResponse.tokenResponse)
+            case MockAPIEndpoint.userinfo.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.userinfo.url, statusCode: 200, headers: MockResponse.headers), MockResponse.userinfoResponse)
+            case MockAPIEndpoint.customHTMLTemplate.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.customHTMLTemplate.url, statusCode: 200, headers: MockResponse.customHTMLTemplateHeaders), MockResponse.customHTMLTemplate)
+            case MockAPIEndpoint.authorization.url.path:
+                return (try mockResponse(url: MockAPIEndpoint.authorization.url, statusCode: 200, headers: MockResponse.authorizeResponseHeaders), MockResponse.authorizeResponse)
+            default:
+                return (try mockResponse(url: MockAPIEndpoint.discovery.url, statusCode: 500), Data())
+            }
+        }
+        
+        let daVinci = DaVinci.createDaVinci { config in
+            config.httpClient = MockURLProtocol.makeClient()
+            
+            config.module(PingDavinci.OidcModule.config) { oidcValue in
+                oidcValue.clientId = self.testClientId
+                oidcValue.scopes = Set(self.testScopes)
+                oidcValue.redirectUri = self.testRedirectUri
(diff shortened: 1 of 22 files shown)


Oidc/Oidc/OidcClient.swift
```diff
@@ -35,8 +35,32 @@ public class OidcClient {
...
+            throw OidcError.networkError(message: "URL not found")
+        }
+        
+        return url
+    }
+    
+    /// OidcClient generateAuthorizeUrl with PAR support.
+    /// When PAR is enabled in the configuration, authorization parameters are pushed to the server before generating the URL.
+    /// - Parameter customParams: Custom parameters to include in the authorization request.
+    public func generateAuthorizeUrl(customParams: [String: String]? = nil) async throws -> URL {
```
vahancouver, line 66: so now we have regular and async versions of the same method (exact same signature). Is the other one being deprecated (if so, we should mark it such). If no, then should we have different names? Even though swift allows this, but it looks confusing...

vahancouver: Also where is this used? I can't seem to find it being called anywhere in the repo.

Oidc/Oidc/OidcClient.swift
```diff
@@ -306,56 +330,146 @@ public class OidcClient {
...
+            // Standard flow: all params on the authorization URL
+            request.url = openId?.authorizationEndpoint ?? ""
+            if !responseMode.isEmpty {
+                request.setParameter(name: OidcClient.Constants.response_mode, value: responseMode)
+            }
+            buildAuthorizeParams(pkce: pkce) { key, value in
+                request.setParameter(name: key, value: value)
+            }
+        }
+        return request
```
vahancouver, line 481: Should we call async (the standard) `populateRequest` here? seems like duplicated code.

Journey/Journey/Module/Agent.swift
```diff
@@ -64,28 +64,14 @@ internal final class CreateAgent: Agent, Sendable {
...
-        guard let openId = config.openId else {
-            throw OidcError.unknown(message: "OpenID configuration not found")
-        }
-        
-        let params: [String: String] = [
-            JourneyConstants.response_type: JourneyConstants.code,
-            JourneyConstants.redirect_uri: config.redirectUri,
-            JourneyConstants.client_id: config.clientId,
-            JourneyConstants.scope: config.scopes.joined(separator: " "),
-            JourneyConstants.state: pkce.state,
```
vahancouver, line 76: Seems like after changes, `pkce.state` is not used anymore.. can you double check this is intentional?

Oidc/Oidc/OidcClient.swift
```diff
@@ -306,56 +330,146 @@ public class OidcClient {
...
+    ///   - pkce: PKCE parameters for enhanced security.
+    ///   - responseMode: The response mode to use.
+    /// - Returns: The populated request ready for execution.
+    public func populateRequest(
+        request: Request,
+        pkce: Pkce,
+        responseMode: String = "pi.flow"
+    ) async throws -> Request {
+        if par, let parEndpoint = openId?.pushedAuthorizationRequestEndpoint {
+            // PAR flow: POST all params to PAR endpoint
```
vahancouver, line 440: should we log which flow is being used?

vahancouver (approved):
Approved with some comments and questions

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: This is a public API so I wanted to make sure this is not a breaking change. I don't think we need to deprecate as the one is async and the other isn't. As Swift allows method overriding I find this OK as we make this more flexible to use the one you want

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: Do you mean if you did PAR or not? I don't see a value given that the developer would have to manage this in the configuration

george-bafaloukas-forgerock: Also you can see what happened in the network logs

Journey/Journey/Module/Agent.swift
george-bafaloukas-forgerock: This is not necessarily needed but it's a good catch that we changes the functionality for no real reason. I will re-introduce it

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: Improved

witrisna (approved):

Oidc/Oidc/OidcClient.swift
```diff
@@ -306,54 +330,146 @@ public class OidcClient {
...
+    ///
+    /// - Parameters:
+    ///   - request: The request to populate.
+    ///   - pkce: PKCE parameters for enhanced security.
+    ///   - responseMode: The response mode to use.
+    /// - Returns: The populated request ready for execution.
+    public func populateRequest(
+        request: Request,
+        pkce: Pkce,
+        responseMode: String = "pi.flow"
```
rodrigoareis:
```suggestion
        responseMode: String = OidcClient.Constants.piflow
```

Oidc/Oidc/OidcClient.swift
```diff
@@ -306,54 +330,146 @@ public class OidcClient {
...
+    ///
+    /// **Standard Flow:** Builds authorization URL with all parameters in the query string.
+    /// **PAR Flow:** POSTs parameters to the PAR endpoint, then uses the returned `request_uri` in the authorization request.
+    ///
+    /// - Parameters:
+    ///   - request: The request to populate.
+    ///   - pkce: PKCE parameters for enhanced security.
+    ///   - responseMode: The response mode to use.
+    /// - Returns: The populated request ready for execution.
+    public func populateRequest(
```
rodrigoareis, line 434: Should this async version be `internal` as well?

Davinci/DavinciTests/DaVinciPARTests.swift
```diff
@@ -0,0 +1,280 @@
...
+import XCTest
+import PingDavinciPlugin
+@testable import PingOrchestrate
+@testable import PingLogger
+@testable import PingOidc
+@testable import PingStorage
+@testable import PingDavinci
+@testable import PingNetwork
+
+final class DaVinciPARTests: XCTestCase, @unchecked Sendable {
```
rodrigoareis, line 20: There are several force-unwraps (!) in this test file (e.g., `request.url!.path`, `node as! ContinueNode`, `userToken!`). While force-unwraps are more acceptable in tests, consider using `XCTUnwrap` for safer test assertions.

Oidc/Oidc/OidcClient.swift
```diff
@@ -35,8 +35,32 @@ public class OidcClient {
             throw OidcError.networkError(message: "HTTP client not found")
         }
         var request = httpClient.request()
-        self.pkce = Pkce.generate()
-        request = config.populateRequest(request: request, pkce: pkce!, responseMode: OidcClient.Constants.query)
+        let generatedPkce = Pkce.generate()
```
rodrigoareis, line 48: I recommend adding a clear doc comment to this synchronous version warning that it does NOT support PAR.

Oidc/Oidc/OidcClient.swift
```diff
@@ -306,54 +330,146 @@ public class OidcClient {
...
+            let immutableParams = formParams
+            let response = try await httpClient.request { req in
+                req.url = parEndpoint
+                req.form(parameters: immutableParams)
+            }
+            guard response.status.isSuccess() else {
+                throw OidcError.apiError(code: response.status, message: "Failed to create PAR request: \(response.bodyAsString())")
+            }
+            
+            let json = try JSONSerialization.jsonObject(with: response.body ?? Data()) as? [String: Any] ?? [:]
```
rodrigoareis: When `response.body` is `nil`, this passes empty `Data()` to `JSONSerialization`, which will throw a serialization error. The `try` will propagate this error, which is fine, but the error message won't be descriptive. Consider adding an explicit guard:
```
guard let responseBody = response.body else {
    throw OidcError.authorizeError(message: "PAR response body is empty")
}
let json = try JSONSerialization.jsonObject(with: responseBody) as? [String: Any] ?? [:]
```

rodrigoareis (commented):
Overall changes looks good. Left some comments

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: made the extension public and removed internal duplicate from Davinci module.

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: We need this public so that it can be used from the Agent of the Journey module

witrisna (approved):

rodrigoareis (approved):
LGTM

PR #154, opened 2026-04-29
Add SingleCheckboxCollector to Davinci, add unit tests

vahancouver:
[SDKS-4920](https://pingidentity.atlassian.net/browse/SDKS-4920) [iOS] [DaVinci] - Support Single Checkbox

- This also includes support for `richContent` in  `SingleCheckboxCollector`
Simulator Screenshot - iPhone Air - 2026-04-29 at 11 20 30
Simulator Screenshot - iPhone Air - 2026-04-29 at 10 53 58

### DIFF
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -95,7 +95,7 @@ extension DaVinci {
                 return QRCodeCollector(with: json)
             })
             await CollectorFactory.shared.register(type: Constants.BOOLEAN, closure: { json in
-                return SingleCheckboxCollector(with: json)
+                return BooleanCollector(with: json)
             })
             if let c: NSObject.Type = NSClassFromString("PingProtect.ProtectCollector") as? NSObject.Type {
                 c.perform(Selector(("registerCollector")))
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -94,6 +94,9 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.QR_CODE, closure: { json in
                 return QRCodeCollector(with: json)
             })
+            await CollectorFactory.shared.register(type: Constants.BOOLEAN, closure: { json in
+                return SingleCheckboxCollector(with: json)
+            })
             if let c: NSObject.Type = NSClassFromString("PingProtect.ProtectCollector") as? NSObject.Type {
                 c.perform(Selector(("registerCollector")))
             }
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -99,6 +99,9 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.BOOLEAN, closure: { json in
                 return BooleanCollector(with: json)
             })
+            await CollectorFactory.shared.register(type: Constants.READ_ONLY_TEXT, closure: { json in
+                return ReadOnlyTextCollector(with: json)
+            })
             if let c: NSObject.Type = NSClassFromString("PingProtect.ProtectCollector") as? NSObject.Type {
                 c.perform(Selector(("registerCollector")))
             }
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -94,6 +94,8 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.QR_CODE, closure: { json in
                 return QRCodeCollector(with: json)
             })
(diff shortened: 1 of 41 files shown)


Davinci/Davinci/collector/BooleanCollector.swift
```diff
@@ -0,0 +1,116 @@
...
+    public required init(with json: [String: Any]) {
+        if let appearanceString = json[Constants.appearance] as? String,
+           let parsedAppearance = SingleCheckboxAppearance(rawValue: appearanceString) {
+            self.appearance = parsedAppearance
+        } else {
+            self.appearance = .checkbox
+        }
+        self.errorMessage = json[Constants.errorMessage] as? String
+        
+        if let richContentDict = json[Constants.richContent] as? [String: Any],
```
vibhorgoswami, line 41: If RichContent is null, are we just reusing the label instead?

vibhorgoswami (approved):
Just one comment, rest LGTM. Android is also aligned.

spetrov (approved):
LGTM! 👍🏻 
I pushed a [commit](https://github.com/ForgeRock/ping-ios-sdk/pull/154/changes/8f6e22a6905ba5f5d21884ce005eaef1151ba6fb) that fixes the [FormFieldsTests](https://github.com/ForgeRock/ping-ios-sdk/pull/154/changes#diff-4bc865cd73aa6fee17ce663c621664c9a6dc774b8c18d319d56359a3481571c6) and adds e2e coverage...

Davinci/Davinci/collector/BooleanCollector.swift
witrisna: I am ok to just leave it null, make it consistent to Android and LabelCollector.

witrisna (approved):

SampleApps/PingExample/PingExample/Collectors/BooleanCollectorView.swift
```diff
@@ -0,0 +1,157 @@
...
+            if newValue {
+                isValid = field.validate().isEmpty
+            }
+        }
+    }
+    
+    // MARK: - Checkbox Appearance
+    
+    private var checkboxAppearance: some View {
+        Button(action: toggleValue) {
```
george-bafaloukas-forgerock: This wraps the entire horizontal stack (including the text label) in a Button. Because the labelContent can contain an AttributedString with interactive .link attributes, wrapping it in a Button creates a known SwiftUI gesture conflict. Tapping the link will likely trigger the button's toggleValue action and swallow the link tap, preventing the URL from opening.

Davinci/Davinci/DaVinci.swift
```diff
@@ -94,6 +94,9 @@ extension DaVinci {
             await CollectorFactory.shared.register(type: Constants.QR_CODE, closure: { json in
                 return QRCodeCollector(with: json)
             })
+            await CollectorFactory.shared.register(type: Constants.BOOLEAN, closure: { json in
```
george-bafaloukas-forgerock, line 99: Can we add some comments here specifying that this actually registers the Boolean collector that can have different types?

george-bafaloukas-forgerock (approved):
LGTM, try to address these issues/comments first

Davinci/Davinci/collector/BooleanCollector.swift
vahancouver: @vibhorgoswami @witrisna I think we should leave it to the developer to check if richContent is nil and use the label property instead if they want to. This way we convey the json as it comes instead of  assigning `label` property's value to the `content` inside `RichContent`. 
Also I see Android is using `richText` property name inside `RichContent`... we don't really have a `richText` inside JSON, it's called `content`. I'm using the name `content` in iOS as it is in the JSON. I think Android should align as well.

SampleApps/PingExample/PingExample/Collectors/BooleanCollectorView.swift
vahancouver: fixed

Davinci/Davinci/DaVinci.swift
vahancouver: Added a comment, also mentioned other similar Collectors that use similar approach

george-bafaloukas-forgerock (approved):

Davinci/Davinci/collector/BooleanCollector.swift
vibhorgoswami: Yes, it's a local variable which is inside the `RichContent` data class. It simplified the actual usage for me and that's why I named it `richText`. Also, since the `RichContent` class is common between `BooleanCollector` and `LabelCollector`, the `LabelCollector` class already has `content` variable which may cause ambiguity when using it.

PR #155, opened 2026-04-30
SDKS-5015 Fixing PIN verification

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-5015) "When registering a new PIN for device binding the second (verify) prompt accepts any PIN entered"

# Description

// SecItemCopyMatching only returns a key reference — the .applicationPassword
                // ACL is evaluated lazily when the key is used. Force that evaluation here by
                // performing a throwaway signature with the user-supplied PIN context, so a
                // wrong PIN fails immediately instead of appearing to succeed.

# Checklist:


vahancouver (approved):
LGTM

spetrov (approved):
LGTM! 👍🏻

PR #156, opened 2026-05-06
SDKS-4924 Add Logger to FIDO module                 

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-4924) "Add Logger to FIDO module                 "

# Description

- Adds `PingLogger` as a direct SPM dependency of `PingFido` (the `.podspec` already declared it; this brings SPM into parity).
  - Routes FIDO ceremony logs through the workflow logger by adding a `logger:` parameter to `Fido.register` and `Fido.authenticate`. Per-call parameter (not a stored property) so concurrent ceremonies on the shared `Fido.shared` instance do not clobber each other's logger.                                                                                                                                                                  
  - The four DaVinci collectors and Journey callbacks (`FidoRegistrationCollector`, `FidoAuthenticationCollector`, `FidoRegistrationCallback`, `FidoAuthenticationCallback`) pass their workflow logger (`davinci?.config.logger` / `journey?.config.logger`) into `Fido` so registration/authentication state transitions and errors emit through the same logger as the surrounding flow.
  - No credential bytes, key material, or challenge data are logged — only state transitions, request counts, and error descriptions.      

# Checklist:

### DIFF
--- a/Fido/Fido/Davinci/AbstractFidoCollector.swift
+++ b/Fido/Fido/Davinci/AbstractFidoCollector.swift
@@ -2,7 +2,7 @@
 //  AbstractFidoCollector.swift
 //  Fido
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
--- a/Fido/Fido/Davinci/AbstractFidoCollector.swift
+++ b/Fido/Fido/Davinci/AbstractFidoCollector.swift
@@ -54,8 +54,8 @@ public class AbstractFidoCollector: AnyFieldCollector, DaVinciAware, Submittable
     public var davinci: DaVinci?
     
     /// The logger for recording Fido related events.
-    public var logger: Logger? {
-        return davinci?.config.logger
+    public var logger: Logger {
+        return davinci?.config.logger ?? LogManager.logger
     }
     
     /// Private storage for the Fido instance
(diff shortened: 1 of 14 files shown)


Package.swift
```diff
@@ -266,6 +266,7 @@ let package = Package(
         .target(
             name: "PingFido",
             dependencies: [
+                "PingLogger",
```
spetrov, line 269: `PingCommons` already brings `PingLogger` dependency transitively, so the direct `PingLogger` dependency seems to be redundant...

spetrov (commented):
Looks good to me, but throughout the SDK Logger is non-optional and falls back to `LogManager.logger` (e.g. `WorkflowConfig`, `HttpClientConfig`, `ReCaptchaEnterpriseCallback`, etc...). 
`PingFido` uses Logger? everywhere instead, which forces logger?.d(...) call sites and silently drops logs when journey/davinci is nil. Was this intentional, or should it align with the rest of the SDK?

Package.swift
george-bafaloukas-forgerock: Technically SPM can resolve transitive modules, but explicitly listing a direct dependency you actively use is the right practice and what SPM's own guidance recommends. Removing it and relying on the PingCommons → PingLogger chain is fragile: if PingCommons ever drops or restructures that dependency, PingFido silently breaks.

george-bafaloukas-forgerock: I am adding this to the podspec as well

george-bafaloukas-forgerock:
> Looks good to me, but throughout the SDK Logger is non-optional and falls back to `LogManager.logger` (e.g. `WorkflowConfig`, `HttpClientConfig`, `ReCaptchaEnterpriseCallback`, etc...). `PingFido` uses Logger? everywhere instead, which forces logger?.d(...) call sites and silently drops logs when journey/davinci is nil. Was this intentional, or should it align with the rest of the SDK?

Updated to match the pattern of other modules

Package.swift
spetrov: Makes sense - thanks for clarifying!

spetrov (approved):

Fido/Fido/Davinci/AbstractFidoCollector.swift
```diff
@@ -54,8 +54,8 @@ public class AbstractFidoCollector: AnyFieldCollector, DaVinciAware, Submittable
     public var davinci: DaVinci?
     
     /// The logger for recording Fido related events.
-    public var logger: Logger? {
-        return davinci?.config.logger
+    public var logger: Logger {
```
rodrigoareis, line 57: Update copyright

Fido/Fido/Journey/FidoCallback.swift
```diff
@@ -28,8 +28,8 @@ public class FidoCallback: AbstractCallback, JourneyAware, ContinueNodeAware, @u
     public var journey: Journey?
     
     /// Logger instance for this callback, obtained from the workflow configuration.
-    public var logger: Logger? {
-        return journey?.config.logger
+    public var logger: Logger {
```
rodrigoareis, line 31: Update copyright

rodrigoareis (approved):
Changes looks good to me. Just minor comments to address

Fido/Fido/Davinci/AbstractFidoCollector.swift
george-bafaloukas-forgerock: resolved

Fido/Fido/Journey/FidoCallback.swift
george-bafaloukas-forgerock: resolved

vahancouver (approved):
LGTM

PR #157, opened 2026-05-08
SDKS-4916 Fix swift build failures — platform bump and if canImport guards

rodrigoareis:
# JIRA Ticket

SDKS-4916 Fix swift build failures — platform bump and if canImport guards
                                                                                                                                                                                       
  ## Summary                                                                                                                                                                                                      
                                                                                                                                                                                                                  
  - Bump `macOS(.v10_15)` → `macOS(.v13)` in `Package.swift` — resolves `Locale.language.languageCode` availability and satisfies GoogleSignIn's macOS minimum requirement (cannot remove macOS entirely due to   
  SPM platform compatibility validation)                                                                                                                                                                          
  - Add `CACHING_GUIDE.md` to `PingStorage` `exclude:` in `Package.swift` — eliminates the SPM "unhandled file" warning                                                                                           
  - Add `condition: .when(platforms: [.iOS])` to `RecaptchaEnterprise`, `PingOneSignals`, and `FacebookLogin` product dependencies — prevents SPM from resolving iOS-only XCFrameworks during native macOS builds 
  - Guard all affected source files with `#if canImport(X)` — allows the macOS build path to compile cleanly without affecting iOS/xcodebuild builds                                          
                                                                                                                                                                                                                  
  ## Root Cause                                                                                                                                                                                                   
                                                                                                                                                                                                                  
  `Package.swift` declares `macOS(.v13)`, so `swift build` on a macOS host compiles all targets for native macOS. Three external XCFramework dependencies (`RecaptchaEnterpriseSDK`, `PingOneSignals`,            
  `FBSDKLoginKit`) only ship iOS slices — causing "no such module" errors. UIKit-dependent source code in TamperDetector, Browser, ExternalIdP, and ExternalIdPGoogle also needed guards.                       
                                                                                                                                                                                                                  
  ## Test Plan                                                                                                                                                                                                    
   
  - [x] `rm -rf .build && swift build` — completes with zero errors                                                                                                                                               
  - [x] `swift build --target PingReCaptchaEnterprise`, `PingProtect`, `PingExternalIdPFacebook`, `PingPush`, `PingDavinci`, `PingOidc`, `PingBrowser`— all pass  
  - [x] all test passed

### DIFF
--- a/DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
+++ b/DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
@@ -105,6 +105,7 @@ public struct HardwareInfo: Codable, Sendable {
             "orientation": isPortrait ? 1 : 0
         ]
         #else
+        // macOS: build target only for 3rd-party library compatibility — not a supported runtime platform
         return [:]
         #endif
     }
@@ -139,6 +140,7 @@ public struct HardwareInfo: Codable, Sendable {
         let cameraCount = discoverySession.devices.count
         return ["numberOfCameras": cameraCount]
         #else
+        // macOS: build target only for 3rd-party library compatibility — not a supported runtime platform
         return [:]
         #endif
     }
--- a/DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
+++ b/DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
@@ -2,15 +2,17 @@
 //  HardwareCollector.swift
 //  DeviceProfile
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
 //
 
 import Foundation
-import UIKit
 import AVFoundation
+#if canImport(UIKit)
+import UIKit
+#endif
 
 // MARK: - HardwareCollector
 
@@ -93,14 +95,18 @@ public struct HardwareInfo: Codable, Sendable {
     /// - Values represent logical screen dimensions
     @MainActor
     private static func getDisplayInfo() -> [String: Int] {
+        #if canImport(UIKit)
         let screenBounds = UIScreen.main.bounds
         let isPortrait = UIDevice.current.orientation.isPortrait
-        
+
         return [
             "width": Int(screenBounds.width),
             "height": Int(screenBounds.height),
             "orientation": isPortrait ? 1 : 0
         ]
+        #else
+        return [:]
+        #endif
     }
     
     /// Retrieves camera system information
@@ -119,6 +125,7 @@ public struct HardwareInfo: Codable, Sendable {
     /// - May return 0 if camera access is restricted
     /// - Includes specialty cameras (telephoto, ultra-wide, etc.)
     private static func getCameraInfo() -> [String: Int] {
+        #if canImport(UIKit)
         let discoverySession = AVCaptureDevice.DiscoverySession(
             deviceTypes: [
                 .builtInTelephotoCamera,
@@ -128,8 +135,11 @@ public struct HardwareInfo: Codable, Sendable {
             mediaType: .video,
             position: .unspecified
         )
-        
+
         let cameraCount = discoverySession.devices.count
         return ["numberOfCameras": cameraCount]
+        #else
+        return [:]
+        #endif
     }
 }
--- a/DeviceProfile/DeviceProfile/Collectors/PlatformCollector.swift
+++ b/DeviceProfile/DeviceProfile/Collectors/PlatformCollector.swift
@@ -2,15 +2,17 @@
 //  PlatformCollector.swift
 //  DeviceProfile
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
 //
 
 import Foundation
-import UIKit
 import PingTamperDetector
+#if canImport(UIKit)
+import UIKit
+#endif
 
 // MARK: - PlatformCollector
 
@@ -72,15 +74,22 @@ public struct PlatformInfo: Codable, Sendable {
     
     /// Initializes platform information by collecting system details
     init() async {
+        #if canImport(UIKit)
         self.platform = await UIDevice.current.systemName
         self.version = await UIDevice.current.systemVersion
         self.device = await UIDevice.current.model
         self.deviceName = await UIDevice.current.name
+        #else
+        self.platform = "macOS"
+        self.version = ProcessInfo.processInfo.operatingSystemVersionString
+        self.device = "Mac"
+        self.deviceName = Host.current().localizedName ?? "Mac"
+        #endif
         self.model = Self.getDeviceModel()
         self.brand = "Apple"
         self.locale = Locale.current.language.languageCode?.identifier
         self.timeZone = TimeZone.current.identifier
-        
+
         self.jailBreakScore = await TamperDetector().analyze()
     }
     
--- a/DeviceProfile/DeviceProfile/Collectors/PlatformCollector.swift
+++ b/DeviceProfile/DeviceProfile/Collectors/PlatformCollector.swift
@@ -81,7 +81,8 @@ public struct PlatformInfo: Codable, Sendable {
         self.deviceName = await UIDevice.current.name
         #else
         self.platform = "macOS"
-        self.version = ProcessInfo.processInfo.operatingSystemVersionString
+        let v = ProcessInfo.processInfo.operatingSystemVersion
+        self.version = "\(v.majorVersion).\(v.minorVersion).\(v.patchVersion)"
         self.device = "Mac"
         self.deviceName = Host.current().localizedName ?? "Mac"
         #endif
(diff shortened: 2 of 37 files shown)


Fido/Fido/Fido.swift
```diff
@@ -153,10 +155,12 @@ public class Fido: NSObject, ASAuthorizationControllerDelegate, ASAuthorizationC
     ///- Returns: The presentation anchor.
     public func presentationAnchor(for controller: ASAuthorizationController) -> ASPresentationAnchor {
         guard let window = window else {
+            #if canImport(UIKit)
             if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
                let window = windowScene.windows.first {
                 return window
             }
+            #endif
             fatalError("Window not set. This should never occur.")
```
george-bafaloukas-forgerock, line 198: Since the SDK only declares macOS as a build target for 3rd-party library compatibility, consumers on macOS should never reach this code path. However the `fatalError` message is now misleading — on macOS the UIKit fallback block is compiled out entirely, so there is genuinely no recovery path. Consider either adding a comment that clarifies why this is safe (`// macOS: build target only — this path is unreachable in practice`), or adding a macOS-specific fallback via `NSApplication.shared.keyWindow` to make the crash truly unreachable:

```swift
#if canImport(UIKit)
if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
   let window = windowScene.windows.first {
    return window
}
#else
if let window = NSApplication.shared.keyWindow {
    return window
}
#endif
fatalError("Window not set. This should never occur.")
```

DeviceProfile/DeviceProfile/LocationManager.swift
```diff
@@ -249,27 +249,40 @@ public class LocationManager: NSObject, ObservableObject, @unchecked Sendable {
...
         case .denied:
             throw LocationError.authorizationDenied
-            
+
         case .restricted:
             throw LocationError.authorizationRestricted
-            
-        @unknown default:
+
+        default:
```
george-bafaloukas-forgerock: The original code used `@unknown default`, which makes the compiler emit a warning if Apple adds a new `CLAuthorizationStatus` case in a future OS. This refactor replaces it with a plain `default`, silently swallowing any future enum cases (mapping them to `authorizationDenied` with no logging). The split into `#if canImport(UIKit)` for `authorizedWhenInUse` above is correct, but the `switch` itself should restore exhaustiveness protection:

```swift
@unknown default:
    throw LocationError.authorizationDenied
```

DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
```diff
@@ -93,14 +95,18 @@ public struct HardwareInfo: Codable, Sendable {
...
         let isPortrait = UIDevice.current.orientation.isPortrait
-        
+
         return [
             "width": Int(screenBounds.width),
             "height": Int(screenBounds.height),
             "orientation": isPortrait ? 1 : 0
         ]
+        #else
+        return [:]
```
george-bafaloukas-forgerock, line 109: The macOS `#else` path returns `[:]` (empty dictionary) for display info. Since macOS is a build-only target for 3rd-party library compatibility and this code won't run in production, the empty return is fine — but a comment would help future readers distinguish intent from omission:

```swift
#else
// macOS: build target only for 3rd-party library compatibility — not a supported runtime platform
return [:]
#endif
```

DeviceProfile/DeviceProfile/Collectors/HardwareCollector.swift
```diff
@@ -128,8 +135,11 @@ public struct HardwareInfo: Codable, Sendable {
             mediaType: .video,
             position: .unspecified
         )
-        
+
         let cameraCount = discoverySession.devices.count
         return ["numberOfCameras": cameraCount]
+        #else
+        return [:]
```
george-bafaloukas-forgerock, line 144: Same as the display info fallback above — macOS is a build-only target so the empty return is intentional. Worth adding the same comment here for consistency:

```swift
#else
// macOS: build target only for 3rd-party library compatibility — not a supported runtime platform
return [:]
#endif
```

DeviceProfile/DeviceProfile/Collectors/TelephonyCollector.swift
```diff
@@ -86,6 +87,10 @@ public struct TelephonyInfo: Codable, Sendable {
             self.carrierName = DeviceProfileConstants.unknown
             self.networkCountryIso = DeviceProfileConstants.unknown
         }
+        #else
+        self.carrierName = DeviceProfileConstants.unknown
```
george-bafaloukas-forgerock, line 92: Since macOS is a build-only target and this code won't run in production, returning `DeviceProfileConstants.unknown` is acceptable. A brief comment would clarify that this is intentional (not a missing implementation), and flag that `nil` would be more semantically accurate if macOS ever gets promoted to a real target — telephony doesn't exist on Mac at all, whereas `"Unknown"` implies data was collected but unreadable:

```swift
#else
// macOS: build target only — telephony not available on this platform
self.carrierName = DeviceProfileConstants.unknown
self.networkCountryIso = DeviceProfileConstants.unknown
#endif
```

DeviceProfile/DeviceProfile/Collectors/PlatformCollector.swift
```diff
@@ -72,15 +74,22 @@ public struct PlatformInfo: Codable, Sendable {
...
     /// Initializes platform information by collecting system details
     init() async {
+        #if canImport(UIKit)
         self.platform = await UIDevice.current.systemName
         self.version = await UIDevice.current.systemVersion
         self.device = await UIDevice.current.model
         self.deviceName = await UIDevice.current.name
+        #else
+        self.platform = "macOS"
+        self.version = ProcessInfo.processInfo.operatingSystemVersionString
```
george-bafaloukas-forgerock: `ProcessInfo.processInfo.operatingSystemVersionString` returns a verbose string like `"Version 13.0 (Build 22A380)"`, while the iOS path returns `"16.0"` via `UIDevice.current.systemVersion`. Since macOS is a build-only target this won't affect production, but if macOS is ever promoted or this data is sent to a server, the structural difference will break any version string parsing. Worth aligning the format now:

```swift
let v = ProcessInfo.processInfo.operatingSystemVersion
self.version = "\(v.majorVersion).\(v.minorVersion).\(v.patchVersion)"
```

ExternalIdP/ExternalIdP/IdpCollector.swift
```diff
@@ -200,6 +201,9 @@ open class IdpCollector: NSObject, Collector, ContinueNodeAware, RequestIntercep
         } catch {
             return .failure(.idpCanceledException(message: error.localizedDescription))
         }
+        #else
+        return .failure(.illegalStateException(message: "Browser authentication is not supported on this platform"))
```
george-bafaloukas-forgerock, line 206: Since macOS is a build-only target and browser-based IdP is iOS-only, returning a failure here is correct and intentional. Worth adding a comment so it's obvious this isn't a gap to fill:

```swift
#else
// macOS: build target only — browser-based IdP authentication requires UIKit and is not a supported use case
return .failure(.illegalStateException(message: "Browser authentication is not supported on this platform"))
#endif
```

george-bafaloukas-forgerock:
**Package.swift — `GoogleSignIn` linked unconditionally on macOS**

`RecaptchaEnterprise`, `PingOneSignals`, and `FacebookLogin` all received `.when(platforms: [.iOS])` conditions because they are iOS-only XCFrameworks. `GoogleSignIn` 9.0.0 does ship a macOS binary so it won't cause a link error — but since every source file in `PingExternalIdPGoogle` is wrapped in `#if canImport(UIKit)`, the framework is linked into the macOS build target while none of its API is actually called. For consistency, and to guard against a future `GoogleSignIn` version dropping macOS support, consider:

```swift
.product(name: "GoogleSignIn", package: "GoogleSignIn-iOS", condition: .when(platforms: [.iOS]))
```

DeviceProfile/DeviceProfile/LocationManager.swift
```diff
@@ -249,26 +249,47 @@ public class LocationManager: NSObject, ObservableObject, @unchecked Sendable {
...
+            statusAuthorized = (status == .authorizedAlways)
+            #endif
+            if statusAuthorized {
                 return try await fetchLocationWithAuthorization()
             } else {
                 throw authorizationErrorForStatus(status)
             }
-            
+
+        case .authorizedAlways:
```
george-bafaloukas-forgerock: Hey, small one here — this `case .authorizedAlways:` arm is actually unreachable 🙂

The `isAuthorized` pre-check just above already captures `.authorizedAlways` on both UIKit and non-UIKit paths, so the `if isAuthorized { return }` guard always short-circuits before execution reaches this switch arm.

Suggest removing the case and adding a brief comment explaining that the switch only handles the remaining states (`.notDetermined`, `.denied`, `.restricted`, `@unknown default`). That way the intent is immediately obvious to anyone reading it later.

george-bafaloukas-forgerock (commented):
Hey, noticed a small inconsistency in `authorizationStatusString` (line 159) — the `case .authorizedWhenInUse:` arm has no `#if canImport(UIKit)` guard, so on macOS it returns `"authorizedWhenInUse"`. But in `requestLocation()`, `.authorizedWhenInUse` is absent from both the `isAuthorized` pre-check and the switch on non-UIKit platforms, so the same status would fall through to `@unknown default` and throw `.authorizationDenied`.

The two methods currently disagree about whether `.authorizedWhenInUse` is valid on macOS. Since macOS is a build-only target it won't bite in production, but it could be a footgun if that policy ever changes. An easy fix would be to wrap the case in `#if canImport(UIKit)` in `authorizationStatusString` too, to keep both methods consistent 🙂

DeviceProfile/DeviceProfile/LocationManager.swift
rodrigoareis: We need to keep `case .authorizedAlways:` rather than deleting it because Swift's exhaustiveness checker for `@objc` enums requires all named cases to be listed. `@unknown default` alone doesn't suppress the warning. I added comments to the code.

george-bafaloukas-forgerock (approved):
LGTM, great work Rodrigo. Just a minor conflict to resolve and it is good to merge

spetrov (approved):
LGTM!

PR #158, opened 2026-05-08
Add richtext support to LabelCollector

vahancouver:
[SDKS-4245](https://pingidentity.atlassian.net/browse/SDKS-4245) [iOS] [DaVinci] [Unified] - Support links coming from Translatable Rich Text (Forms)


vibhorgoswami (approved):
LGTM

SampleApps/PingExample/PingExample/Collectors/RichTextBuilder.swift
```diff
@@ -0,0 +1,54 @@
...
+//  Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import SwiftUI
+import PingDavinci
+
+func buildAttributedString(from richContent: RichContent) -> AttributedString {
```
george-bafaloukas-forgerock: `buildAttributedString(from:)` is declared as a module-level free function. In Swift, free functions in the same module share a flat namespace — any other file in `PingExample` could shadow or conflict with this name, and it won't surface cleanly in autocomplete. Since this sample app is shown to customers as reference code, scope it inside a dedicated type instead:

```swift
enum RichTextBuilder {
    static func build(from richContent: RichContent) -> AttributedString { … }
}
```

SampleApps/PingExample/PingExample/Collectors/RichTextBuilder.swift
```diff
@@ -0,0 +1,54 @@
...
+        guard let closeRange = remaining[afterOpen...].range(of: "}}") else {
+            result.append(AttributedString(String(remaining[openRange.lowerBound...])))
+            remaining = remaining[remaining.endIndex...]
+            break
+        }
+
+        let placeholderKey = String(remaining[afterOpen..<closeRange.lowerBound])
+
+        if let replacement = richContent.replacements[placeholderKey] {
+            if replacement.type == "link", let href = replacement.href, let url = URL(string: href) {
```
george-bafaloukas-forgerock: `URL(string: href)` returns `nil` for malformed server-supplied URLs (relative paths, non-RFC-3986 characters, etc.). When it does, the guard silently falls through to plain unlinked text with no logging or indication to the developer. For a reference sample app, at minimum add a comment explaining the fallback, or add an `assertionFailure` in debug builds so the issue is visible during integration testing.

Davinci/Davinci/RichContent.swift
```diff
@@ -0,0 +1,29 @@
...
+    public let target: String?
+}
+
+/// Rich content for a form field, enabling template-based text with embedded links.
+/// - `content`: A template string with `{{placeholder}}` tokens.
+/// - `replacements`: A dictionary mapping placeholder keys to their replacement details.
+public struct RichContent: Sendable {
+    public let content: String
+    public let replacements: [String: RichContentReplacement]
+}
```
george-bafaloukas-forgerock: Both `RichContent` and `RichContentReplacement` are pure value types with no stored closures or references, so `Equatable` conformance is trivial and free. Without it, callers can't use `==` or `XCTAssertEqual` directly on these structs — they have to compare individual fields, as the tests currently do. Add `Equatable` at the definition site:

```swift
public struct RichContentReplacement: Sendable, Equatable { … }
public struct RichContent: Sendable, Equatable { … }
```

george-bafaloukas-forgerock (changes requested):
FYI Trying the polaris:review skill and auto submitting. We can discuss and will do a manual review as well. The 2 comments on the code seem valid on a first read
## Overall

Functionally solid — the JSON parsing is correct, the test coverage is thorough, and the type extraction to a shared file is the right call. Three things need to be addressed before merge: the CHANGELOG entry (required by Jira AC), `Equatable` conformance on the new public types, and the global function scope in the sample app.

**Missing CHANGELOG entry** — The Jira AC explicitly requires documentation. `RichContent`, `RichContentReplacement`, and `LabelCollector.richContent` are public API additions that belong in the CHANGELOG under the current version section, with the Jira reference (SDKS-4245).

**Parsing duplication** — The `richContent` JSON parsing block in `LabelCollector.init` is byte-for-byte identical to the same block in `BooleanCollector`. If the `richContent` JSON schema ever grows a new field, both files need updating in sync. Consider a `static func parse(from:) -> RichContent?` on `RichContent` itself to centralise this logic.

**`testRichContentTypeIsSharedBetweenCollectors`** — Add a brief comment clarifying what "shared" means here (both collectors accept the same JSON shape, not a shared instance). The name is slightly misleading — a future reader might expect the test to assert identity rather than structural equivalence.

Davinci/Davinci/RichContent.swift
spetrov, line 1: `RichContent.swift` is placed at the module root (Davinci/Davinci/) while all other collector-supporting types live in `Davinci/Davinci/collector/`. Consider moving the file...

Davinci/Davinci/collector/LabelCollector.swift
```diff
@@ -24,12 +24,31 @@ public class LabelCollector: Collector, @unchecked Sendable {
     public private(set) var key: String = ""
     /// The label content.
     public private(set) var content: String = ""
-    
+    /// Optional rich content with template text and link replacements.
+    public private(set) var richContent: RichContent?
```
spetrov, line 28: Update Davinci readme file to include `.richContent` too in the [LabelCollector (LABEL)](https://github.com/ForgeRock/ping-ios-sdk/tree/polaris/SDKS-4245-support-links-translatable-rich-text/Davinci#labelcollector-label) section

spetrov (approved):
LGTM. 
Just 2 minor suggestions from me.

SampleApps/PingExample/PingExample/Collectors/RichTextBuilder.swift
vahancouver: Addressed

Davinci/Davinci/RichContent.swift
vahancouver: Addressed

Davinci/Davinci/collector/LabelCollector.swift
vahancouver: Added

Davinci/Davinci/RichContent.swift
vahancouver: moved

george-bafaloukas-forgerock (approved):
LGTM

PR #160, opened 2026-05-25
Implement OAuth 2.0 Device Authorization Grant (RFC 8628) support

vahancouver:
[SDKS-4785](https://pingidentity.atlassian.net/browse/SDKS-4785) [Ping SDKs] [iOS]: Update OIDC Module to Support Device Authorization Grant (RFC 8628)
[SDKS-5027](https://pingidentity.atlassian.net/browse/SDKS-5027) [iOS] Update Development Sample App to support Device Authorization Grant.

  - Add OidcDeviceClient with polling loop and DeviceAuthorizationResponse/DeviceFlowStatus models
  - Add OidcClientConfig and OpenIdConfiguration fields for device authorization endpoint
  - Add DaVinci+Device.swift entry point and update Davinci Oidc/Request modules
  - Update Journey Oidc module with device approval flow support
  - Add JourneyPlugin constants for device authorization
  - Add BrowserLauncher changes for device flow
  - Add DeviceFlowView and ApproveDeviceView UI with QR scanner to sample app
  - Update ConfigurationManager, ViewModels, and ContentView for device flow configuration
  - Add unit tests for OidcDeviceClient (core + additional) and DaVinci/Journey device flows
  - Update Oidc, Davinci, and Journey READMEs with device grant usage

### DIFF
--- a/Browser/Browser/BrowserLauncher.swift
+++ b/Browser/Browser/BrowserLauncher.swift
@@ -285,7 +285,9 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
             throw BrowserError.externalUserAgentFailure
         }
         var presentingVC = root
-        while let presented = presentingVC.presentedViewController {
+        while let presented = presentingVC.presentedViewController,
+              !presented.isBeingDismissed,
+              !(presented is UIAlertController) {
             presentingVC = presented
         }
         
--- a/Browser/Browser/BrowserLauncher.swift
+++ b/Browser/Browser/BrowserLauncher.swift
@@ -279,10 +279,14 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
         safariVC.modalPresentationStyle = .fullScreen
         
         guard let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
-              let presentingVC = windowScene.windows.first(where: { $0.isKeyWindow })?.rootViewController else {
+              let root = windowScene.windows.first(where: { $0.isKeyWindow })?.rootViewController else {
             logger.e("Fail to launch SFSafariViewController; missing presenting ViewController", error: nil)
             state = .idle
             throw BrowserError.externalUserAgentFailure
+        }
+        var presentingVC = root
+        while let presented = presentingVC.presentedViewController {
+            presentingVC = presented
         }
         
         // We set state BEFORE presenting to prevent race conditions
--- a/Davinci/Davinci/module/Oidc.swift
+++ b/Davinci/Davinci/module/Oidc.swift
@@ -48,9 +48,12 @@ public class OidcModule {
                let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
                let userCode = components.queryItems?.first(where: { $0.name == "user_code" })?.value,
                !userCode.isEmpty {
+                // Move key from sharedContext (workflow-scoped) into flowContext (this invocation only)
+                // so success can detect the device-flow path, and a subsequent start() is unaffected.
                 _ = daVinciFlow.sharedContext.removeValue(forKey: SharedContext.Keys.daVinciVerificationUriCompleteKey)
+                context.flowContext.set(key: SharedContext.Keys.daVinciVerificationUriCompleteKey, value: uriString)
                 config.logger.d( "Oidc: device code completion flow detected, skipping authorization request")
-                return config.populateDeviceFlowVerificationRequest(request: request, userCode: userCode)
+                return try config.populateDeviceFlowVerificationRequest(request: request, userCode: userCode)
             }
             
             
@@ -66,8 +69,9 @@ public class OidcModule {
         setup.success { @Sendable context, success in
             
             // Device-code completion flow: token exchange already handled externally, skip.
-            if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String {
-               return SuccessNode(input: success.input, session: success.session)
+            // The key was moved into flowContext by start; sharedContext key was already consumed there.
+            if context.flowContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) != nil {
+                return SuccessNode(input: success.input, session: success.session)
             }
             
             let cloneConfig: OidcClientConfig = config.clone()
--- a/Davinci/Davinci/module/Oidc.swift
+++ b/Davinci/Davinci/module/Oidc.swift
@@ -46,7 +46,7 @@ public class OidcModule {
             if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String,
                let url = URL(string: uriString),
                let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
-               let userCode = components.queryItems?.first(where: { $0.name == "user_code" })?.value,
+               let userCode = components.queryItems?.first(where: { $0.name == OidcClient.Constants.userCodeSnake })?.value,
                !userCode.isEmpty {
                 // Move key from sharedContext (workflow-scoped) into flowContext (this invocation only)
                 // so success can detect the device-flow path, and a subsequent start() is unaffected.
--- a/Davinci/Davinci/module/Oidc.swift
+++ b/Davinci/Davinci/module/Oidc.swift
@@ -13,6 +13,7 @@ import Foundation
 import PingOidc
 import PingOrchestrate
 import PingNetwork
+import PingLogger
 
 /// A module that integrates OIDC capabilities into the DaVinci workflow.
 public class OidcModule {
@@ -40,6 +41,19 @@ public class OidcModule {
         
         // Starts the module.
         setup.start { @Sendable context, request in
+            
+            // Check for Device Flow
+            if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String,
+               let url = URL(string: uriString),
+               let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
+               let userCode = components.queryItems?.first(where: { $0.name == "user_code" })?.value,
+               !userCode.isEmpty {
+                _ = daVinciFlow.sharedContext.removeValue(forKey: SharedContext.Keys.daVinciVerificationUriCompleteKey)
+                config.logger.d( "Oidc: device code completion flow detected, skipping authorization request")
+                return config.populateDeviceFlowVerificationRequest(request: request, userCode: userCode)
+            }
+            
+            
             // When user starts the flow again, revoke previous token if exists
             await daVinciFlow.daVinciUser()?.revoke()
             
@@ -50,6 +64,12 @@ public class OidcModule {
         
         // Handles success of the module.
         setup.success { @Sendable context, success in
+            
+            // Device-code completion flow: token exchange already handled externally, skip.
+            if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String {
+               return SuccessNode(input: success.input, session: success.session)
+            }
+            
             let cloneConfig: OidcClientConfig = config.clone()
             
             let flowPkce = context.flowContext.get(key: SharedContext.Keys.pkceKey) as? Pkce
@@ -87,4 +107,8 @@ extension SharedContext.Keys {
     
     /// The key used to store the OIDC client configuration in the shared context.
     public static let oidcClientConfigKey = "com.pingidentity.davinci.OidcClientConfig"
+    
+    /// The shared context key under which the `verificationUriComplete` URL string is stored
+    /// for the DaVinci device authorization flow.
+    public static let daVinciVerificationUriCompleteKey = "com.pingidentity.davinci.VerificationUriComplete"
 }
--- a/Davinci/Davinci/module/Request.swift
+++ b/Davinci/Davinci/module/Request.swift
@@ -54,19 +54,14 @@ extension OidcClientConfig {
         }
 
         // Strip "/as/device_authorization" to obtain the tenant-scoped base URL.
-        let baseUrl = deviceAuthEndpoint.hasSuffix(Constants.asDeviceAuthorizationPath)
-            ? String(deviceAuthEndpoint.dropLast(Constants.asDeviceAuthorizationPath.count))
+        let baseUrl = deviceAuthEndpoint.hasSuffix(OidcClient.Constants.asDeviceAuthorizationPath)
+            ? String(deviceAuthEndpoint.dropLast(OidcClient.Constants.asDeviceAuthorizationPath.count))
             : deviceAuthEndpoint
 
         request.url = "\(baseUrl)/applications/\(clientId)/deviceFlow"
-        request.setParameter(name: Constants.userCodeCamel, value: userCode)
+        request.setParameter(name: OidcClient.Constants.userCodeCamel, value: userCode)
 
         return request
-    }
-
-    private enum Constants {
-        static let asDeviceAuthorizationPath = "/as/device_authorization"
-        static let userCodeCamel = "userCode"
     }
 }
 
--- a/Davinci/Davinci/module/Request.swift
+++ b/Davinci/Davinci/module/Request.swift
@@ -24,5 +24,46 @@ extension OidcClientConfig {
     ) async throws -> Request {
         return try await populateRequest(request: request, pkce: pkce, responseMode: OidcClient.Constants.piflow)
     }
+    
+    /// Populates a request to verify a user code in the Device Authorization Grant flow (RFC 8628).
+    ///
+    /// Constructs the device-flow verification URL from
+    /// `OpenIdConfiguration.deviceAuthorizationEndpoint` by stripping the trailing
+    /// `/as/device_authorization` segment to obtain the tenant-scoped base URL, then appending
+    /// `/applications/{clientId}/deviceFlow`. The user code is sent as a `userCode` query parameter
+    /// (PingOne format).
+    ///
+    /// **Example**
+    /// - Endpoint: `https://auth.pingone.ca/{tenantId}/as/device_authorization`
+    /// - Result:   `https://auth.pingone.ca/{tenantId}/applications/{clientId}/deviceFlow?userCode={userCode}`
+    ///
+    /// - Parameters:
+    ///   - request: The request to populate.
+    ///   - userCode: The user code obtained from the device authorization response that needs to be verified.
+    /// - Returns: The populated `Request` ready for execution.
+    internal func populateDeviceFlowVerificationRequest(
+        request: Request,
+        userCode: String
+    ) -> Request {
+        let deviceAuthEndpoint = openId?.deviceAuthorizationEndpoint ?? ""
+
+        // Strip "/as/device_authorization" to obtain the tenant-scoped base URL.
+        let baseUrl: String
+        if deviceAuthEndpoint.hasSuffix(Constants.asDeviceAuthorizationPath) {
+            baseUrl = String(deviceAuthEndpoint.dropLast(Constants.asDeviceAuthorizationPath.count))
+        } else {
+            baseUrl = deviceAuthEndpoint
+        }
+
+        request.url = "\(baseUrl)/applications/\(clientId)/deviceFlow"
+        request.setParameter(name: Constants.userCodeCamel, value: userCode)
+
+        return request
+    }
+
+    private enum Constants {
+        static let asDeviceAuthorizationPath = "/as/device_authorization"
+        static let userCodeCamel = "userCode"
+    }
 }
 
--- a/Davinci/Davinci/module/Request.swift
+++ b/Davinci/Davinci/module/Request.swift
@@ -33,6 +33,9 @@ extension OidcClientConfig {
     /// `/applications/{clientId}/deviceFlow`. The user code is sent as a `userCode` query parameter
     /// (PingOne format).
     ///
+    /// PingOne format paths look like `/{tenantId}/as/device_authorization`, whereas
+    /// custom-domain paths look like `/as/device_authorization`.
+    ///
     /// **Example**
     /// - Endpoint: `https://auth.pingone.ca/{tenantId}/as/device_authorization`
     /// - Result:   `https://auth.pingone.ca/{tenantId}/applications/{clientId}/deviceFlow?userCode={userCode}`
@@ -41,19 +44,19 @@ extension OidcClientConfig {
     ///   - request: The request to populate.
     ///   - userCode: The user code obtained from the device authorization response that needs to be verified.
     /// - Returns: The populated `Request` ready for execution.
+    /// - Throws: `OidcError.unknown` if the device authorization endpoint is not available.
     internal func populateDeviceFlowVerificationRequest(
         request: Request,
         userCode: String
-    ) -> Request {
-        let deviceAuthEndpoint = openId?.deviceAuthorizationEndpoint ?? ""
+    ) throws -> Request {
+        guard let deviceAuthEndpoint = openId?.deviceAuthorizationEndpoint, !deviceAuthEndpoint.isEmpty else {
+            throw OidcError.unknown(message: "Device authorization endpoint not available")
+        }
 
         // Strip "/as/device_authorization" to obtain the tenant-scoped base URL.
-        let baseUrl: String
-        if deviceAuthEndpoint.hasSuffix(Constants.asDeviceAuthorizationPath) {
-            baseUrl = String(deviceAuthEndpoint.dropLast(Constants.asDeviceAuthorizationPath.count))
-        } else {
-            baseUrl = deviceAuthEndpoint
-        }
+        let baseUrl = deviceAuthEndpoint.hasSuffix(Constants.asDeviceAuthorizationPath)
+            ? String(deviceAuthEndpoint.dropLast(Constants.asDeviceAuthorizationPath.count))
+            : deviceAuthEndpoint
 
         request.url = "\(baseUrl)/applications/\(clientId)/deviceFlow"
         request.setParameter(name: Constants.userCodeCamel, value: userCode)
--- a/Davinci/DavinciTests/DaVinciDeviceTests.swift
+++ b/Davinci/DavinciTests/DaVinciDeviceTests.swift
@@ -1,0 +1,227 @@
+//
+//  DaVinciDeviceTests.swift
+//  DavinciTests
+//
+//  Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import XCTest
+@testable import PingOrchestrate
+@testable import PingLogger
+@testable import PingOidc
+@testable import PingStorage
+@testable import PingDavinci
+@testable import PingNetwork
+
+// MARK: - populateDeviceFlowVerificationRequest tests
+
+/// Tests for `OidcClientConfig.populateDeviceFlowVerificationRequest`.
+final class PopulateDeviceFlowVerificationRequestTests: XCTestCase {
+
+    private func makeConfig(deviceAuthEndpoint: String, clientId: String = "my-client") -> OidcClientConfig {
+        let config = OidcClientConfig()
+        config.clientId = clientId
+        config.openId = OpenIdConfiguration(
+            authorizationEndpoint: "https://auth.example.com/authorize",
+            tokenEndpoint: "https://auth.example.com/token",
+            userinfoEndpoint: "https://auth.example.com/userinfo",
+            endSessionEndpoint: "https://auth.example.com/signoff",
+            revocationEndpoint: "https://auth.example.com/revoke",
+            deviceAuthorizationEndpoint: deviceAuthEndpoint
+        )
+        return config
+    }
+
+    // MARK: - Standard PingOne endpoint strips /as/device_authorization suffix
+
+    func testStripsAsDeviceAuthorizationSuffix() {
+        let endpoint = "https://auth.pingone.ca/tenant123/as/device_authorization"
+        let config = makeConfig(deviceAuthEndpoint: endpoint)
+        let request = URLSessionHttpRequest()
+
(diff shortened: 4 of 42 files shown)


Davinci/Davinci/module/Oidc.swift
```diff
@@ -50,6 +64,12 @@ public class OidcModule {
         
         // Handles success of the module.
         setup.success { @Sendable context, success in
+            
+            // Device-code completion flow: token exchange already handled externally, skip.
+            if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String {
```
george-bafaloukas-forgerock: Hey — I think this guard is unreachable. Up in `setup.start` (line 51) we already do `daVinciFlow.sharedContext.removeValue(forKey: ...daVinciVerificationUriCompleteKey)`, so by the time `setup.success` runs the key is gone and this `if let` is always false. We then fall through to the regular OIDC path and create a `CreateAgent` / `OidcUser` / `UserDelegate` against the device-flow session — exactly what the comment says we want to skip. The unit test `testVerificationUriCompleteKeyIsConsumedAfterFirstStart` actually confirms the key is consumed in `start`, which doubles as evidence this branch never fires.

A couple of options:
- Don't remove the key in `start`, and instead clear it after `success` runs.
- Or stash a separate `isDeviceFlow` flag in `sharedContext` that `success` can read.

Whichever feels cleaner. 🙂

Oidc/Oidc/OidcDeviceClient.swift
```diff
@@ -0,0 +1,295 @@
...
+                                )
+                            )
+                            return
+                        }
+
+                    } catch let urlError as URLError {
+                        // Network/timeout error — apply exponential backoff and continue
+                        consecutiveTimeouts += 1
+                        let backoffMultiplier = 1 << min(consecutiveTimeouts, 3)
+                        interval = min(max(interval * backoffMultiplier, 5), 60)
```
george-bafaloukas-forgerock: Small one: once we back off here, `interval` is mutated permanently — even after the network recovers and we go back to seeing `authorization_pending`, we keep polling at the backed-off value (potentially 60s) for the rest of the flow. RFC 8628 §3.5 says the client SHOULD use the server-provided `interval` and only increase on `slow_down`.

Would it be worth keeping the original interval in a separate `let` and restoring it once `consecutiveTimeouts == 0`? The existing URLError test breaks out of the loop on the first `.polling`, so it doesn't catch this today.

Oidc/Oidc/OidcDeviceClient.swift
```diff
@@ -0,0 +1,295 @@
...
+
+        // Capture everything needed inside the stream closure — device_code never touches logger
+        let deviceCode = deviceAuthResponse.deviceCode
+        let tokenEndpoint = openId.tokenEndpoint
+        let storage = config.storage
+        let capturedConfig = config
+        let capturedLogger = logger
+
+        return AsyncThrowingStream { continuation in
+            let task = Task.detached {
```
george-bafaloukas-forgerock: Heads up — `Task.detached` doesn't inherit cancellation from the caller. We do wire `continuation.onTermination = { _ in task.cancel() }` below, which covers the case where the consumer iterates and then drops the stream. But if a caller holds the stream without ever calling `next` (or holds it past the lifetime of the calling task), the detached task happily polls forever.

A plain `Task { ... }` would inherit cancellation automatically. If there's a reason we need detached here, a comment explaining the cancellation contract would help future readers.

Oidc/Oidc/OidcDeviceClient.swift
```diff
@@ -0,0 +1,295 @@
...
+    /// (either via the redirect callback or because the user closed it); the polling loop
+    /// in `deviceAuthorization()` continues independently while this method is awaiting.
+    ///
+    /// - Parameter verificationUriComplete: The verification URI (including `user_code`)
+    ///   returned in the `DeviceAuthorizationResponse`.
+    /// - Throws: `BrowserError.externalUserAgentCancelled` if the user closed the browser
+    ///   before completing the flow, or any other error surfaced by `BrowserLauncher`.
+    ///   No-ops (returns without throwing) if `verificationUriComplete` is not a valid URL
+    ///   or `config.redirectUri` has no scheme.
+    public func authorize(verificationUriComplete: String) async throws {
```
george-bafaloukas-forgerock, line 255: Friendly footgun warning: if `verificationUriComplete` doesn't parse as a URL, or `redirectUri` has no scheme, this method just logs and returns — no throw, no signal to the caller. Production callers will see "the function returned successfully" but the browser never opened, and they'll wonder why polling never resolves.

Would it be reasonable to throw a typed error here (or return a `Result`) so misconfiguration fails loudly? Documented in the docstring, but still surprising at runtime.

Oidc/Oidc/OidcDeviceClient.swift
```diff
@@ -0,0 +1,295 @@
...
+
+/// A client that implements the RFC 8628 Device Authorization Grant (requesting-device side).
+///
+/// Use `OidcDeviceClient` when the device cannot open a browser directly
+/// (e.g., a smart TV or CLI tool). Call `deviceAuthorization()` to start the
+/// flow and receive a stream of `DeviceFlowStatus` updates. The stream yields
+/// `.started` with the `userCode` and `verificationUri` to display, then
+/// `.polling` on each poll, and finally `.success`, `.accessDenied`, or
+/// `.expired` when the flow resolves.
+public class OidcDeviceClient: @unchecked Sendable {
```
george-bafaloukas-forgerock, line 27: Not introduced by this PR, but worth flagging since this PR materially expands the surface area: we mark this `@unchecked Sendable` while it stores a `private let config: OidcClientConfig` — and `OidcClientConfig` is itself `@unchecked Sendable` with mutable `var` fields (`openId`, `httpClient`, `storage`, …). The polling closure captures `let capturedConfig = config`, and the same instance is reachable from any actor with no synchronization.

Not necessarily blocking, but as we lean further into Swift Concurrency this is a latent race waiting to bite. Either lock around the mutable bits or document a clear "don't mutate after first use" contract on `OidcClientConfig`.

Davinci/Davinci/module/Request.swift
```diff
@@ -24,5 +24,46 @@ extension OidcClientConfig {
...
+
+        // Strip "/as/device_authorization" to obtain the tenant-scoped base URL.
+        let baseUrl: String
+        if deviceAuthEndpoint.hasSuffix(Constants.asDeviceAuthorizationPath) {
+            baseUrl = String(deviceAuthEndpoint.dropLast(Constants.asDeviceAuthorizationPath.count))
+        } else {
+            baseUrl = deviceAuthEndpoint
+        }
+
+        request.url = "\(baseUrl)/applications/\(clientId)/deviceFlow"
```
george-bafaloukas-forgerock, line 61: Two concerns here, both around fail-loud behavior:

1. The `else` branch on line 55 keeps `baseUrl = deviceAuthEndpoint` when the endpoint *doesn't* end in `/as/device_authorization`. For any non-PingOne AS, that produces a URL like `https://generic.example.com/oauth/device_authorization/applications/{clientId}/deviceFlow` — which will silently 404 instead of telling us we're in the wrong code path.
2. When `deviceAuthorizationEndpoint` is empty (nil from discovery), we end up with `/applications/{clientId}/deviceFlow` — a path with no scheme/host — and the test `testEmptyEndpointProducesApplicationsPath` documents this rather than rejecting it.

Could we throw / fail-fast when the PingOne suffix is absent or the endpoint is empty? Right now the SDK silently builds bad URLs in those cases.

Journey/Journey/Module/Oidc.swift
```diff
@@ -39,16 +41,45 @@ public class OidcModule {
...
+                            request.setHeader(name: approvalConfig?.cookie ?? JourneyConstants.cookie, value: ssoToken.value)
+                            request.form(parameters: [
+                                JourneyConstants.userCode: userCode,
+                                JourneyConstants.decision: JourneyConstants.decisionAllow,
+                                JourneyConstants.csrf: ssoToken.value
+                            ])
+                        }
+                        journeyFlow.config.logger.i("Oidc: device approval response status: \(response.status)")
+                    } catch {
+                        journeyFlow.config.logger.e("Oidc: failed to submit approval to AM device endpoint", error: error)
```
george-bafaloukas-forgerock: Tiny UX concern: when the approval POST throws, we log it and continue, and then below we still return a `SuccessNode`. So the *approving* device's UI happily shows "approved!" while the *requesting* device just keeps polling until it times out — which is going to look like a broken flow to a user.

Would it make sense to surface this either as a `FailureNode`, or via a status field on the `SuccessNode`'s session, so the caller can show the right state?

SampleApps/PingExample/PingExample/DeviceFlowView.swift
```diff
@@ -0,0 +1,437 @@
...
+        Task { try? await client.authorize(verificationUriComplete: uriComplete) }
+    }
+
+    func hasExistingUser() async -> Bool {
+        guard let client = ConfigurationManager.shared.deviceClient else { return false }
+        return await client.user() != nil
+    }
+
+    func reset() {
+        streamTask?.cancel()
```
george-bafaloukas-forgerock, line 96: Small nit on the sample app: `streamTask?.cancel()` here signals cancellation but doesn't `await streamTask?.value`, so the in-flight poll inside `OidcDeviceClient` may still be running when a new flow is started. Combined with `Task.detached` over in the client (separate comment), it's possible to end up with two flows in flight.

Not critical for sample-app code, but if folks copy this into their own apps it'll quietly bite them.

Browser/Browser/BrowserLauncher.swift
```diff
@@ -279,11 +279,15 @@ public final class BrowserLauncher: NSObject, BrowserLauncherProtocol {
...
-              let presentingVC = windowScene.windows.first(where: { $0.isKeyWindow })?.rootViewController else {
+              let root = windowScene.windows.first(where: { $0.isKeyWindow })?.rootViewController else {
             logger.e("Fail to launch SFSafariViewController; missing presenting ViewController", error: nil)
             state = .idle
             throw BrowserError.externalUserAgentFailure
         }
+        var presentingVC = root
+        while let presented = presentingVC.presentedViewController {
+            presentingVC = presented
+        }
```
george-bafaloukas-forgerock, line 295: Drilling to the topmost presented VC is the right move 👍 — one defensive thought: if any intermediate `presentedViewController` is mid-dismissal (or is a `UIAlertController`, which can't itself present another VC), `presentingVC.present(safariVC, animated: true)` on line 294 will log the classic "presenting view controller is already presenting" warning or silently no-op.

A `presentingVC.isBeingDismissed == false` guard inside the loop, plus skipping `UIAlertController`, would harden this nicely.

Oidc/OidcTests/OidcDeviceClientTests.swift
```diff
@@ -0,0 +1,609 @@
...
+                XCTFail("Unexpected .failure — stream should have continued after URLError: \(error)")
+            }
+        }
+
+        XCTAssertTrue(statuses.contains("started"), "Stream should have yielded .started")
+        XCTAssertTrue(statuses.contains("polling"), "Stream should have yielded .polling after URLError backoff")
+        // After first URLError, consecutiveTimeouts=1, backoffMultiplier=2, interval=min(0*2,60)=0; but interval
+        // cannot go below the original value — verify it is at least 0 (not negative) and the stream continued
+        XCTAssertNotNil(pollingIntervalAfterError, "Should have captured the backoff polling interval")
+        XCTAssertGreaterThanOrEqual(pollingIntervalAfterError ?? -1, 0, "Backoff interval should be non-negative")
```
george-bafaloukas-forgerock: This assertion is essentially "`Int >= 0`" — the comment on line 595 even admits it. It doesn't really verify the backoff actually backed off.

If we go with the suggestion to reset the interval after recovery (the separate comment on `OidcDeviceClient.swift:203`), the assertion here could become something like:
- `interval` after URLError is strictly greater than the original, **and**
- after the next successful `authorization_pending`, `interval` is back to the original.

That would actually pin the contract we want.

Davinci/DavinciTests/DaVinciDeviceTests.swift
```diff
@@ -0,0 +1,227 @@
...
+    // MARK: - Empty deviceAuthorizationEndpoint produces a well-formed (if empty-base) URL
+
+    func testEmptyEndpointProducesApplicationsPath() {
+        let config = makeConfig(deviceAuthEndpoint: "", clientId: "c1")
+        let request = URLSessionHttpRequest()
+
+        let populated = config.populateDeviceFlowVerificationRequest(request: request, userCode: "XX")
+
+        XCTAssertTrue(populated.url?.contains("/applications/c1/deviceFlow") == true,
+                      "Even with an empty endpoint the applications path should be appended")
```
george-bafaloukas-forgerock: This test asserts that an empty `deviceAuthorizationEndpoint` produces a path-only URL — which is exactly the buggy behavior flagged on `Request.swift:58`. We probably want the test to encode the contract we *want* (e.g. "empty endpoint throws / results in a `FailureNode`") rather than locking in the current behavior.

Worth revisiting once we decide how that path should fail.

george-bafaloukas-forgerock (commented):
Thanks for the work on this — RFC 8628 is a chunky spec to land cleanly. I left a handful of inline comments. The big one is the unreachable guard in `Davinci/module/Oidc.swift` — the `success` block's device-flow check can never trigger because `start` already removed the key. The rest are smaller polish items around backoff, cancellation, and a couple of tests that pin behavior we probably don't actually want to lock in. Happy to chat through any of them.

Journey/Journey/Module/Oidc.swift
```diff
@@ -39,16 +41,45 @@ public class OidcModule {
...
+                   !userCode.isEmpty {
+                    let approvalConfig: JourneyConfig? = journeyFlow.config as? JourneyConfig
+                    do {
+                        let response = try await journeyFlow.config.httpClient.request { request in
+                            request.url = uriString
+                            request.setHeader(name: approvalConfig?.cookie ?? JourneyConstants.cookie, value: ssoToken.value)
+                            request.form(parameters: [
+                                JourneyConstants.userCode: userCode,
+                                JourneyConstants.decision: JourneyConstants.decisionAllow,
+                                JourneyConstants.csrf: ssoToken.value
```
george-bafaloukas-forgerock: Quick question on this. Is the CSRF value expected to be the SSO token from AM in all cases?

Oidc/Oidc/OidcDeviceClient.swift
witrisna: **Android baseline (`OidcDeviceClient.kt`):** Android does not apply exponential back-off on network errors; any exception propagates to the `catch (e: Exception)` block which emits `DeviceFlowStatus.Failure`.

Journey/Journey/Module/Oidc.swift
witrisna: [Alignment with Android] **POST failure is silently swallowed — approving device sees success while the requesting device keeps polling until expiry.**

When the POST to `verificationUriComplete` throws (line 56–65), the error is caught, logged, and `SuccessNode` is returned anyway. The approving user sees the Journey as completed, but the requesting device never receives the approval signal.

**Android baseline (`journey/module/Oidc.kt`):** Android checks `response.status.isSuccess()` and throws `ApiException` on non-2xx:
```kotlin
if (!response.status.isSuccess()) {
    logger.w("OidcDevice: device-user POST returned status ${response.status}…")
    throw ApiException(response.status, response.body())
}
```
This propagates the error to the Journey caller so it can be surfaced as a `FailureNode`.

**Suggested fix:** Align with Android — check the response status after the POST and return a `FailureNode` (or rethrow) on non-2xx instead of silently succeeding.

Journey/Journey/Module/Oidc.swift
```diff
@@ -39,16 +41,45 @@ public class OidcModule {
         
         // Handles success of the module.
         setup.success { @Sendable context, success in
+            // RFC 8628 approving-device: when Journey.start was given a verificationUriComplete,
+            // POST user_code + decision=allow back to that URL with the SSO token as cookie.
+            // No-op when the key is absent.
+            if let uriString = journeyFlow.sharedContext.get(key: SharedContext.Keys.journeyVerificationUriCompleteKey) as? String,
+               !uriString.isEmpty {
+                let existingSsoToken = await !success.session.value.isEmpty ? success.session : journeyFlow.session()
```
witrisna: [Alignment with Android] **Missing final fallback when both `success.session` is empty and `journeyFlow.session()` returns nil.**

```swift
let existingSsoToken = await !success.session.value.isEmpty ? success.session : journeyFlow.session()
```
If both are empty/nil, `existingSsoToken` is `nil`. The `if let ssoToken = existingSsoToken as? SSOToken` guard fails silently, the approval POST is skipped entirely, and `SuccessNode` is returned — the approving device shows success but the requesting device never gets approved.

**Android baseline (`journey/module/Oidc.kt`):**
```kotlin
val existingSession = success.session.takeIf { it.value.isNotEmpty() } ?: journey.session() ?: success.session
```
The fallback ensures there is always attempt the POST with (even an empty session fails loudly on the server).

Oidc/Oidc/OidcClientConfig.swift
```diff
@@ -84,8 +88,12 @@ public class OidcClientConfig: @unchecked Sendable {
         if openId != nil {
             return
         }
-        
+
```
witrisna, line 95: [Alignment with Android] **iOS `openIdOverride` closure vs Android `openId { }` DSL — different capability for fully offline / discovery-free configuration.**

Android PR #199 added an `openId { block }` DSL that sets `OpenIdConfiguration` directly and skips the discovery call entirely:
```kotlin
config.openId {
    deviceAuthorizationEndpoint = "https://…/as/device_authorization"
    tokenEndpoint = "https://…/token"
}
```
iOS `openIdOverride` is applied _after_ `discover()` succeeds, for Android, it has to configure all endpoint instead of particular one, will review if Android should align with iOS on this.

vahancouver: Android has aligned with iOS. no change needed here

Davinci/Davinci/module/Oidc.swift
vahancouver: Good catch — the success branch was unreachable. Took your first option: in start we now move the key from sharedContext (workflow-scoped) into flowContext (per-invocation) instead of removing it. success reads from flowContext, so the device-flow branch fires correctly, and a subsequent start() on the same workflow still sees the sharedContext slot empty. Reused the existing daVinciVerificationUriCompleteKey to avoid introducing a parallel key. The consumed-key test still passes.

Oidc/Oidc/OidcDeviceClient.swift
vahancouver: Fixed. Captured let baseInterval = deviceAuthResponse.interval and reset to it on every successful authorization_pending (which also sets consecutiveTimeouts = 0). Backoff now uses baseInterval * backoffMultiplier so each computation starts from the server-provided interval rather than compounding off a previously-elevated value. Also updated the test per your separate comment.

Oidc/Oidc/OidcDeviceClient.swift
vahancouver: Re-evaluated and you're right — Task { } is the better fit. deviceAuthorization() is a non-isolated async throws and the stream closure is non-isolated, so Task { } runs on the global executor (no MainActor serialization concern). The cancellation contract via continuation.onTermination still works either way. Switched to Task { }.

Oidc/Oidc/OidcDeviceClient.swift
vahancouver: Fixed. authorize(verificationUriComplete:) now throws OidcError.unknown with a descriptive message when the URL fails to parse or  redirectUri has no scheme, so misconfiguration fails loudly at the call site. Updated the docstring to enumerate the throw paths.

Oidc/Oidc/OidcDeviceClient.swift
vahancouver: Took the documentation route for now. Added an - Important: block to both OidcDeviceClient and OidcClientConfig calling out the  "configure before passing, don't mutate after" contract. A proper lock/actor refactor is a larger surface change and out of scope  for this PR — happy to file a follow-up if you'd like.

Davinci/Davinci/module/Request.swift
vahancouver: Aligned with the Android implementation. The function now throws OidcError.unknown when the endpoint is nil/empty. For the  suffix-absent case, we keep the endpoint as base (matches Android's removeSuffix behavior) — non-PingOne ASes will still 404 if  used, but that surfaces as a server-side error rather than a silent local mis-construction. Updated the four happy-path tests with  throws/try and replaced testEmptyEndpointProducesApplicationsPath with testEmptyEndpointThrows + testNilEndpointThrows.

Journey/Journey/Module/Oidc.swift
vahancouver: Aligned with Android. Removed the silent do/catch swallow — the approval POST now throws OidcError.apiError on non-2xx, which  propagates up through next() and is surfaced as a FailureNode so the approving device's UI reflects reality. Test added:  testNon2xxApprovalResponsePropagates.

SampleApps/PingExample/PingExample/DeviceFlowView.swift
vahancouver: Added a comment noting that cancel() only signals cancellation and that production code should await the task before starting a new flow to avoid two concurrent polling loops. Kept the sample minimal so the focus stays on the SDK API.

Browser/Browser/BrowserLauncher.swift
vahancouver: Done. The walk now skips any presentedViewController that's isBeingDismissed or is a UIAlertController, so we won't try to present  on top of a stale or non-presenting controller.

Oidc/OidcTests/OidcDeviceClientTests.swift
vahancouver: Tightened. Now asserts pollingIntervalAfterError == 5 (the floor enforced by min(max(baseInterval * backoffMultiplier, 5), 60) for  the first URLError), pinning the actual contract instead of just >= 0.

Davinci/DavinciTests/DaVinciDeviceTests.swift
vahancouver: Replaced. The test now encodes the contract — testEmptyEndpointThrows and testNilEndpointThrows both assert OidcError.unknown is  thrown, rather than locking in the buggy path-only URL behavior.

Journey/Journey/Module/Oidc.swift
vahancouver: Yes — for AM-backed Journey flows, the SSO token (iPlanetDirectoryPro cookie) is also used as the CSRF token on the approval POST.  AM accepts the SSO token in the csrf form field for cross-site request forgery protection on this endpoint, so the value comes from the same session. This matches the Android implementation.

Oidc/Oidc/OidcDeviceClient.swift
vahancouver: Acknowledged. We're keeping the iOS-side backoff with the fix from George's comment (reset to baseInterval on  authorization_pending, exponential up to 60s on transient URLErrors). Rationale: iOS apps see more flaky cellular conditions where  a short network blip would otherwise terminate the flow. Happy to revisit if you'd prefer strict parity — could swap to "propagate  as .failure on first network error" in a follow-up.

Journey/Journey/Module/Oidc.swift
george-bafaloukas-forgerock: Vahan — the comment says "Align with Android: always attempt the POST", but the current code path doesn't actually do that.

When `success.session.value.isEmpty == true` and `journeyFlow.session()` returns `nil`, the nil-coalescing falls back to `success.session` — the same empty session. The `if let ssoToken = resolvedSession as? SSOToken` guard then fails (it's not an SSOToken), the POST is skipped entirely, and execution falls through to return `SuccessNode`. The approving device shows success; the requesting device keeps polling until expiry.

The Android version you cited uses:
```kotlin
val existingSession = success.session.takeIf { it.value.isNotEmpty() } ?: journey.session() ?: success.session
```
This always assigns *something* to `existingSession` and proceeds to the POST — a bad/empty token surfaces as a server-side 401, not a silent local success.

Can you confirm the intended behaviour here? A few options:

**Option A — match Android exactly**: always attempt the POST; let the server reject a bad token loudly.

**Option B — fail-loud locally**: when both sessions resolve to a non-SSOToken, throw or return `FailureNode` rather than silently returning `SuccessNode`.

**Option C — intentional no-op**: if the design intent is that a non-SSOToken session means "this is not an AM flow, skip the POST", then the current code is correct but `testNoOpWhenSessionIsNotSSOToken` needs a comment explicitly documenting that design decision so future readers don't assume it's a bug.

The test as it stands asserts the silent-success path without any explanation of why that's correct, which makes it look like the unresolved case from the original review.

Davinci/DavinciTests/DaVinciDeviceTests.swift
```diff
@@ -0,0 +1,245 @@
...
+    func testFallbackBaseWhenSuffixAbsent() throws {
+        let endpoint = "https://custom.example.com/device_authorization"
+        let config = makeConfig(deviceAuthEndpoint: endpoint)
+        let request = URLSessionHttpRequest()
+
+        let populated = try config.populateDeviceFlowVerificationRequest(request: request, userCode: "WXYZ-5678")
+
+        XCTAssertEqual(populated.url,
+                       "https://custom.example.com/device_authorization/applications/my-client/deviceFlow?userCode=WXYZ-5678",
+                       "URL should use full endpoint as base when /as/device_authorization suffix is absent")
```
george-bafaloukas-forgerock: This test asserts a URL that will always 404 on any non-PingOne AS — `populateDeviceFlowVerificationRequest` appends the PingOne-specific `/applications/{clientId}/deviceFlow` path to whatever base it derives, so a custom endpoint without the suffix will produce a URL that no standard OAuth AS knows how to handle.

The behaviour itself is acceptable (discussed in the `Request.swift` thread — surfaces as a server-side error rather than a silent local mis-construction). But pinning the exact wrong URL as the assertion makes this look like a verified-correct output. A future caller who points this at a non-PingOne AS will get a 404, but this test will still pass and give no indication why.

Suggestion: either add a `// NOTE: this URL is intentionally PingOne-specific and will 404 on other AS implementations` comment on the assertion, or change the assertion to only verify the function *doesn't throw* (since the real contract for the suffix-absent case is "don't crash, let the server handle it").

SampleApps/PingExample/PingExample/DeviceFlowView.swift
```diff
@@ -0,0 +1,439 @@
...
```
george-bafaloukas-forgerock, line 96: The comment added to `start()` (line 42–43) documents the cancel-without-await limitation there, but `reset()` has the exact same pattern — `streamTask?.cancel()` followed immediately by `streamTask = nil` — and that comment doesn't cover this path.

Anyone reading `reset()` in isolation (or copying it into production code) won't see the caveat. Worth adding the same note here:
```swift
func reset() {
    // cancel() signals cancellation; production code should await the task value before
    // treating the flow as fully stopped to avoid two flows running concurrently.
    streamTask?.cancel()
    streamTask = nil
}
```

Oidc/Oidc/OidcDeviceClient.swift
```diff
@@ -0,0 +1,301 @@
...
+                                consecutiveTimeouts = 0
+                                interval = baseInterval
+                                pollCount += 1
+                                continuation.yield(.polling(
+                                    pollCount: pollCount,
+                                    pollInterval: interval,
+                                    nextPollAt: Date().addingTimeInterval(Double(interval))
+                                ))
+
+                            case Constants.slowDown:
```
george-bafaloukas-forgerock, line 167: The backoff reset fix (capturing `baseInterval` and resetting `interval = baseInterval` on `authorization_pending`) only applies to the URLError recovery path. The `slow_down` case here compounds permanently for the lifetime of the flow: each `slow_down` adds 5 to the current `interval`, and no subsequent response resets it.

Concrete example: server sends `slow_down` three times → `interval` goes 5 → 10 → 15. Then `authorization_pending` arrives — `interval` stays at 15 (the reset is in the URLError branch, not here). RFC 8628 §3.5 says to increase by 5 seconds per `slow_down`, which does mean compounding, but there's no mechanism to ever unwind a `slow_down` storm back to `baseInterval` after conditions improve.

This is essentially the same root issue as George's original backoff comment, applied to `slow_down` instead of URLError. The existing test (`testSlowDownIncreasesInterval`) breaks out after the first `slow_down` and doesn't cover the compound case or verify that a subsequent `authorization_pending` restores the interval.

Suggestion: either reset to `baseInterval` before adding 5 on each `slow_down` (so each one is a fresh +5, not cumulative), or at minimum add a test that sends two consecutive `slow_down`s and then `authorization_pending` and asserts what `interval` should be at each step.

Davinci/Davinci/module/Oidc.swift
```diff
@@ -40,6 +41,22 @@ public class OidcModule {
         
         // Starts the module.
         setup.start { @Sendable context, request in
+            
+            // Check for Device Flow
+            if let uriString = daVinciFlow.sharedContext.get(key: SharedContext.Keys.daVinciVerificationUriCompleteKey) as? String,
+               let url = URL(string: uriString),
+               let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
+               let userCode = components.queryItems?.first(where: { $0.name == "user_code" })?.value,
```
george-bafaloukas-forgerock: Can we move this in a constant as well?

JourneyPlugin/JourneyPlugin/Constants.swift
```diff
@@ -144,4 +144,10 @@ public enum JourneyConstants {
     public static let description = "description"
     public static let submitButtonText = "submitButtonText"
     public static let pageFooter = "pageFooter"
+
+    /// RFC 8628 Device Authorization Grant approval constants
+    public static let userCode = "user_code"
```
george-bafaloukas-forgerock, line 149: To my comment above, this exists here, so please use it on the other file

george-bafaloukas-forgerock (approved):
lgtm

spetrov (approved):
LGTM!

rodrigoareis (approved):
LGTM

PR #161, opened 2026-05-27
Feature/mfa integration

gioraping:
# P14C-82521

[JIRA](https://pingidentity.atlassian.net/browse/P14C-82521) "Add PingOneMFA module to the orchestratio SDK on iOS"

# Description

* New PingOneMFA module — a clean async/await wrapper around the binary PingOneSDK (PingOneSDK XCFramework). Bridges all callback-based PingOneSDK APIs to async throws functions via withCheckedThrowingContinuation, using a @globalActor (PingOneMFAActor) for thread-safe state. Exposed surface: initialize(geo:), setDeviceToken(_:), pair(pairingKey:), getDeviceInfo(), getOneTimePasscode(), processRemoteNotification(userInfo:), processRemoteNotificationAction(identifier:authenticationMethod:userInfo:), generateMobilePayload(), getNotificationCategories().
* Actionable push notifications — registers UNNotificationCategory objects from PingOneSDK so Approve/Deny banner actions work without the app being foregrounded; routes UNUserNotificationCenterDelegate actions through processRemoteNotificationAction.
* PingExample sample app integration — adds a full MFA section: QR-code scanner for device pairing, paired accounts list, live-countdown OTP screen, push notification modal (approve/deny with number-matching support), and mobile payload view.
https://pingidentity.atlassian.net/browse/P14C-86659
* Unit tests — AccountParserTests, PingOneMFATests, PingOneMFAErrorTests, and a MockPingOneMFA covering initialization, pairing, token registration, OTP, push processing, and error bridging.
Package.swift / Package.resolved updates — adds PingOneMFA as a new library target and pulls in the PingOneSDK binary dependency.

# Testing
Set-up native application in PingOne environment

Pairing: tap QR Code Registration → scan a PingOne pairing QR → success dialog → accounts list shows the paired account
OTP: tap One-Time Passcode → code displays with countdown; refreshes automatically at zero
Payload: tap Mobile Payload → payload shown; Copy button writes to clipboard
Push (foreground): trigger a push from PingOne while app is open → approval screen appears; approve/deny shows success dialog
Push (background): trigger a push while app is backgrounded → system banner with Allow/Deny buttons → tapping either completes the request
Cancellation: approve on a second device while the approval screen is open → screen dismisses automatically

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added PingOne MFA support for device pairing, account management, one-time passcodes, mobile payload generation, and push notification approval workflows.
  * Added regional configuration options and structured account, notification, OTP, and error information.
  * Added sample app screens for QR pairing, accounts, OTPs, payloads, and notification actions.
* **Documentation**
  * Added setup instructions, API references, usage examples, and troubleshooting guidance.
* **Tests**
  * Added comprehensive coverage for MFA workflows, errors, concurrency, notifications, and sample integrations.


PingOneMFA/PingOneMFA/PingOneMFA.swift
```diff
@@ -0,0 +1,285 @@
...
+    internal private(set) static var isInitialized: Bool = false
+
+    /// Initializes the PingOneMFA SDK with the provided geographic region.
+    /// This method should be called before using any other methods in the PingOneMFA SDK.
+    /// This method is idempotent — if already initialized, it returns immediately.
+    ///
+    /// - Parameter geo: The geographic region for the PingOneMFA SDK.
+    /// - Throws: `PingOneMFAError` if initialization fails.
+    public nonisolated static func initialize(geo: Geo) async throws {
+        if await isInitialized {
```
george-bafaloukas-forgerock: 🔴 **Idempotency here is racy (TOCTOU) and contradicts the doc comment.**

The guard read (`await isInitialized`) and the write (inside the `configure` callback) are two separate actor hops with an `await` suspension *and* a full network round-trip in between. Two concurrent first-time callers both observe `false`, so both call `PingOne.configure(...)`. The doc says *"if already initialized, it returns immediately"*, but the flag is only set **after** `configure` completes.

The sample app happens to dodge this via `ClientInitializationActor`'s own `isPingOneMFAInitializing` guard, but the README tells direct SDK consumers to just call `initialize()` — and they'd hit the race. Could we move the guard inside the actor (set an `isInitializing` flag synchronously before the `await`, or make this an actor-isolated method rather than `nonisolated`)? Happy to pair on this one.

PingOneMFA/PingOneMFA/PingOneMFA.swift
```diff
@@ -0,0 +1,285 @@
...
+    /// - Returns: An `OtpCodeInfo` containing the current passcode and seconds remaining.
+    /// - Throws: `PingOneMFAError` if the SDK call fails or returns no passcode info.
+    public nonisolated static func getOneTimePasscode() async throws -> OtpCodeInfo {
+        return try await withCheckedThrowingContinuation { continuation in
+            PingOne.getOneTimePasscode { passcodeInfo, error in
+                if let error = error {
+                    continuation.resume(throwing: PingOneMFAError(error))
+                } else if let passcodeInfo = passcodeInfo {
+                    let now = Date().timeIntervalSince1970
+                    let secondsRemaining = Int(passcodeInfo.validUntil - now)
```
george-bafaloukas-forgerock: 🟠 **`secondsRemaining` can come back negative.**

If `validUntil <= now` (an already-expired passcode, or a little device clock skew), `Int(passcodeInfo.validUntil - now)` is negative — and that's an undocumented value handed to every consumer (the API table implies non-negative "seconds until expiry"). I'd clamp at the source here with `max(0, ...)` rather than relying on each UI layer to guard it. (The sample app's `max(1, ...)` clamp then drives a ~1 Hz refetch loop, so fixing it here helps there too.)

PingOneMFA/PingOneMFA/PingOneMFA.swift
```diff
@@ -0,0 +1,285 @@
...
+
+    /// Returns all registered MFA accounts for this device, along with any non-fatal
+    /// diagnostic errors reported by the upstream SDK.
+    ///
+    /// - Returns: A tuple `(accounts, errors)` where `accounts` is the parsed
+    ///   `PingOneMfaAccount` array and `errors` is `nil` if the SDK reported no errors,
+    ///   otherwise an array of `PingOneMFAError` values carrying diagnostic detail.
+    /// - Throws: `PingOneMFAError` when the SDK returns no data and at least one real error,
+    ///   or when the SDK returns neither data nor errors.
+    public nonisolated static func getDeviceInfo() async throws -> (accounts: [PingOneMfaAccount], errors: [PingOneMFAError]?) {
```
george-bafaloukas-forgerock: 🟡 **The `(accounts:errors:)` tuple return is an unusual contract.**

This is the only method that both can `throw` *and* hands back an `errors` array on "success". Every other method either returns a value or throws. A small dedicated result type (or routing diagnostics to a logged side-channel) would read more consistently with the rest of the surface — and it's the reason the README example below is out of sync. Not blocking, just flagging the shape.

PingOneMFA/PingOneMFA/PingOneMFA.swift
```diff
@@ -0,0 +1,285 @@
...
+            pingOneGeo = .Europe
+        case .australia:
+            pingOneGeo = .Australia
+        case .canada:
+            pingOneGeo = .Canada
+        case .singapore:
+            pingOneGeo = .Singapore
+        }
+
+        return try await withCheckedThrowingContinuation { continuation in
```
george-bafaloukas-forgerock: 🟡 _(reuse)_ The same `withCheckedThrowingContinuation { … resume(throwing: PingOneMFAError(error)) … }` shape is repeated across all 8 public methods. A single small generic `bridge` helper would collapse the boilerplate and keep the error-mapping policy single-sourced, so a future change to error handling is one edit instead of eight. Purely a maintainability nicety.

PingOneMFA/PingOneMFA/AccountParser.swift
```diff
@@ -0,0 +1,54 @@
...
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import Foundation
+
+internal struct AccountParser {
+
+    internal static func parse(_ deviceInfo: [String: Any]?) -> [PingOneMfaAccount] {
+        guard let deviceInfo,
+              let data = try? JSONSerialization.data(withJSONObject: deviceInfo),
```
george-bafaloukas-forgerock: 🟠 **Both `try?` here silently swallow failures into an empty list.**

If `deviceInfo` contains a single non-JSON-serializable value (an `NSDate`, an unexpected `NSNull`, a non-`String` nested key), `JSONSerialization.data` throws and the whole account list silently becomes `[]` — indistinguishable from "this user has no accounts." Since this is a boundary against a binary SDK you don't control, could we at least log the failure so it's diagnosable? It also feeds the empty-vs-error ambiguity over in `getDeviceInfo()`.

PingOneMFA/PingOneMFA/PingOneMFAError.swift
```diff
@@ -0,0 +1,70 @@
...
+    init(errors: [Error]) {
+        let internal_ = errors.map { e -> PingOneMFAInternalError in
+            let ns = e as NSError
+            return PingOneMFAInternalError(
+                code: ns.code,
+                message: ns.localizedDescription,
+                userInfo: ns.userInfo.compactMapValues { "\($0)" }
+            )
+        }
+        self.message = internal_.first?.message ?? "Unknown error"
```
george-bafaloukas-forgerock: 🟡 **`init(errors:)` is documented as "aggregating multiple errors", but `.message` only keeps the first.**

`internalErrorsList` preserves the full set (great), but the headline `message`/`errorDescription` that `setDeviceToken` callers see — and that the README prints — drops everything after the first. Could we either join the messages, or tweak the doc to say "first of N; see `internalErrorsList`"?

PingOneMFA/PingOneMFA/PingOneMFAError.swift
```diff
@@ -0,0 +1,70 @@
...
+public struct PingOneMFAError: Error, LocalizedError, Sendable {
+    public let message: String
+
+    /// Structured list of individual SDK errors, or `nil` when the failure did not originate
+    /// from the native SDK (e.g. an unexpected exception).
+    public let internalErrorsList: [PingOneMFAInternalError]?
+
+    public var errorDescription: String? { message }
+
+    init(_ error: Error) {
```
george-bafaloukas-forgerock, line 40: 🟡 **Latent footgun: `init(_ error: Error)` will re-wrap an existing `PingOneMFAError`.**

If a `PingOneMFAError` is ever passed in, it gets bridged through `as NSError` and produces a `"Code=… UserInfo={…}"` message wrapping the original. All current call sites pass native SDK errors so it's not hit today, but a small `if let existing = error as? PingOneMFAError { self = existing; return }` guard would make it refactor-proof.

PingOneMFA/PingOneMFA/PingOneMFAError.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    init(_ message: String = "Unknown error") {
+        self.message = message
+        self.internalErrorsList = nil
+    }
+
+    /// Creates an error aggregating multiple native SDK errors.
+    /// - Parameter errors: The array of native `NSError` values returned by the upstream SDK.
+    init(errors: [Error]) {
+        let internal_ = errors.map { e -> PingOneMFAInternalError in
```
george-bafaloukas-forgerock: 🟢 _(nit)_ The trailing-underscore escape `let internal_` to dodge the keyword reads a bit awkwardly — `let mapped` or `let internalErrors` (matching the `internalErrorsList` property) would be clearer.

PingOneMFA/PingOneMFA/PushNotification.swift
```diff
@@ -0,0 +1,95 @@
...
+//  of the MIT license. See the LICENSE file for details.
+//
+
+import Foundation
+import PingOneSDK
+
+public typealias MFAPushNotification = PushNotification
+
+/// A push notification received from PingOne MFA representing an authentication request.
+public struct PushNotification: @unchecked Sendable, Identifiable {
```
george-bafaloukas-forgerock: 🟡 **`@unchecked Sendable` permanently silences the concurrency checker here.**

It's doing real load-bearing work: `notificationObject` (a non-`Sendable` type from the binary SDK) is read from the public `approve`/`deny`/`pushType`/`getNumbersChallenge` members outside `@PingOneMFAActor`, and this is also posted across `NotificationCenter`. Could we either confine `NotificationObject` access to the actor, or add a short comment documenting *why* this is safe? As written, a future SDK that makes the type properly `Sendable` also wouldn't surface.

PingOneMFA/PingOneMFA/PingOneMfaAccount.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// The cloud region where this account is registered (e.g. `"NA"`, `"EU"`).
+    public let region: String
+    /// The unique identifier for this account.
+    public let id: String
+    /// The identifier of the device this account is bound to.
+    public let deviceId: String
+    /// The PingOne environment this account belongs to.
+    public let environmentId: String
+    /// The display name of the account.
+    public var name: String
```
george-bafaloukas-forgerock: 🟢 _(nit)_ `name` and `family` are `var` while the other four fields are `let`, and nothing mutates an account after parsing. Looks unintentional — making them all `let` would tighten the value type unless mutation is actually expected.

PingOneMFA/PingOneMFA/Geo.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  PingOneMFAConfig.swift
```
george-bafaloukas-forgerock: 🟢 _(nit)_ File header says `PingOneMFAConfig.swift`, but this file is `Geo.swift` — looks like a copy-paste leftover.

PingOneMFA/README.md
```diff
@@ -0,0 +1,353 @@
...
+} catch {
+    print("Pairing failed: \(error.localizedDescription)")
+}
+```
+
+### Retrieve Paired Accounts
+
+```swift
+do {
+    let accounts = try await PingOneMFA.getDeviceInfo()
```
george-bafaloukas-forgerock: 🔴 **This example doesn't compile.** `getDeviceInfo()` returns `(accounts: [PingOneMfaAccount], errors: [PingOneMFAError]?)`, not `[PingOneMfaAccount]`, so `for account in accounts` is iterating a tuple. The API table further down (line ~282) also lists the return as `-> [PingOneMfaAccount]`. Since this is a shipped public doc, a consumer copy-pasting it hits a build error. Could we update both the snippet and the table to the tuple shape?

PingOneMFA/README.md
```diff
@@ -0,0 +1,353 @@
...
+| `processRemoteNotification(userInfo:)` | `async throws -> PushNotification` | Convert an APNS `userInfo` payload to a typed `PushNotification`. |
+| `processRemoteNotificationAction(identifier:authenticationMethod:userInfo:)` | `async throws -> PushNotification?` | Handle a notification banner action; returns `nil` when the SDK handled it internally. |
+| `generateMobilePayload()` | `async throws -> String` | Generate a mobile payload for server-side authentication. |
+| `getNotificationCategories()` | `Set<UNNotificationCategory>` | Return notification categories to register with `UNUserNotificationCenter`. |
+
+### `PingOneMfaAccount`
+
+| Field | Type | Description |
+|---|---|---|
+| `region` | `String` | Region key from the PingOne response (e.g. `"NA"`, `"EU"`) |
```
george-bafaloukas-forgerock: 🟠 **Region example values are wrong.** This says `region` is e.g. `"NA"` / `"EU"`, but `AccountParser` uses the raw top-level JSON keys and the tests assert these are `"NorthAmerica"` / `"Europe"`. Worth correcting so the documented format matches reality.

PingOneMFA/README.md
```diff
@@ -0,0 +1,353 @@
...
+- **Push Notifications (foreground)** — approve or deny incoming authentication requests while the app is active
+- **Mobile Payload** — generate a cryptographic mobile payload for server-side authentication flows
+
+---
+
+## Getting Started
+
+### Prerequisites
+
+- iOS 15 or higher
```
george-bafaloukas-forgerock: 🟡 **iOS version mismatch.** Prereqs say "iOS 15 or higher", but `Package.swift` declares `.iOS(.v16)` — so iOS 15 isn't actually achievable for this package. Should be iOS 16+.

PingOneMFA/README.md
```diff
@@ -0,0 +1,353 @@
...
+| `.singapore` | Singapore |
+
+### 2. Register the APNS Push Token
+
+Call `setDeviceToken()` each time the system delivers a new push token — typically from `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`:
+
+```swift
+func application(_ application: UIApplication,
+                 didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+    Task {
```
george-bafaloukas-forgerock, line 100: 🟡 **Missing precondition.** This snippet calls `setDeviceToken` directly, but the SDK's `setDeviceToken` doesn't self-initialize — it calls straight into `PingOne.setDeviceToken`. The sample app correctly does `ensurePingOneMFAInitialized()` first. Could we note the `initialize()`-first ordering here (the way Troubleshooting does for OTP/payload)?

PingOneMFA/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+The `PingOneMFA` module wraps the PingOne MFA native SDK [PingOneSDK](https://github.com/pingidentity/pingone-mobile-sdk-ios) behind a clean, async/await Swift API. It is the adapter layer between your application and the PingOne MFA platform. All PingOne SDK callbacks are bridged to `async throws` functions — callers never need to work with raw callback closures.
+
+---
+
+## Features
+
+- **Device Pairing** — pair new MFA accounts by scanning a QR code or entering a pairing key
+- **OTP** — retrieve the current one-time passcode and its remaining validity window
+- **Push Notifications (foreground)** — approve or deny incoming authentication requests while the app is active
```
george-bafaloukas-forgerock, line 15: 🟡 _(minor)_ The Features list only mentions "Push Notifications (foreground)", but a headline of this PR is actionable **banner/background** actions (`processRemoteNotificationAction` + `getNotificationCategories`). Worth calling that out here.

PingOneMFA/PingOneMFATests/PingOneMFATests.swift
```diff
@@ -0,0 +1,366 @@
...
+        await PingOneMFA.reset()
+        MockPingOneMFA.reset()
+        try await super.tearDown()
+    }
+
+    // MARK: - Initialization Tests
+
+    /// Mirrors ProtectTests.test04: calls initialize() twice via mock and asserts
+    /// initializeCalled == true after both calls (idempotency guard exercised at mock layer).
+    func test04_InitDoesNotReinitializeIfAlreadyInitialized() async throws {
```
george-bafaloukas-forgerock, line 33: 🟠 **This test's name promises coverage that doesn't exist.** It's called `...DoesNotReinitializeIfAlreadyInitialized` and the comment says "idempotency guard exercised at mock layer", but `MockPingOneMFA.initialize` has no guard — it increments `initializeCallCount` unconditionally, and the test then asserts the count is `2` (the *opposite* of idempotency). The real `PingOneMFA.initialize` idempotency logic is never exercised. Could we either test the real type's idempotency, or rename this so it doesn't imply coverage it doesn't have?

PingOneMFA/PingOneMFATests/PingOneMFATests.swift
```diff
@@ -0,0 +1,366 @@
...
+        } catch {
+            XCTFail("Wrong error type: \(error)")
+        }
+    }
+
+    // MARK: - Thread Safety Tests
+
+    /// Mirrors ProtectTests.test12: fires 5 concurrent initialize() calls via mock,
+    /// asserts mock was called (idempotency under concurrency).
+    func test15_ConcurrentInitializationCalls() async throws {
```
george-bafaloukas-forgerock, line 220: 🟠 **The thread-safety test is itself thread-unsafe.** The 5 concurrent tasks each do `initializeCallCount += 1` on a `nonisolated(unsafe)` static `Int` with no synchronization — a genuine read-modify-write data race. The `XCTAssertEqual(..., 5)` can lose increments and will trip Thread Sanitizer. Could we back the counter with an actor or an `OSAllocatedUnfairLock` so it actually validates concurrency safety?

PingOneMFA/PingOneMFATests/PingOneMFATests.swift
```diff
@@ -0,0 +1,366 @@
...
+
+    func test09_MockGetDeviceInfoHappyPath() async throws {
+        // Given
+        let expectedAccount = PingOneMfaAccount(
+            region: "NorthAmerica",
+            id: "user-id-1",
+            deviceId: "device-id-1",
+            environmentId: "env-id-1",
+            name: "Test User",
+            family: "User",
```
george-bafaloukas-forgerock: 🟢 _(nit)_ Trailing comma after the final argument (`family: "User",`). That only compiles on a Swift 6.1+ toolchain (SE-0439), while `Package.swift` declares `swift-tools-version: 6.0`. Harmless on current Xcode, but inconsistent with the declared tools version — worth normalizing.

PingOneMFA/PingOneMFATests/MockPingOneMFA.swift
```diff
@@ -0,0 +1,151 @@
...
+    nonisolated(unsafe) static var getOneTimePasscodeCalled = false
+    nonisolated(unsafe) static var processPushNotificationCalled = false
+    nonisolated(unsafe) static var getMobilePayloadCalled = false
+    nonisolated(unsafe) static var lastGeo: Geo?
+
+    // Return values for happy-path tests
+    nonisolated(unsafe) static var accountsReturnValue: [PingOneMfaAccount] = []
+    nonisolated(unsafe) static var otpReturnValue = OtpCodeInfo(code: "123456", secondsRemaining: 30)
+    nonisolated(unsafe) static var mobilePayloadReturnValue = "mockMobilePayload"
+    // collectPush cannot return a real PushNotification in tests because NotificationObject
```
george-bafaloukas-forgerock: 🟡 **Push happy-path is entirely untested, which leaves `parseAPNSAlert` at 0% coverage.** The comments are honest that `NotificationObject` can't be constructed in tests — but the one piece of real, pure logic in those methods is `parseAPNSAlert` (title/message extraction). Could we make it `internal` and unit-test it directly against dict-alert / string-alert / missing-alert payloads? That sidesteps the `NotificationObject` problem entirely.

PingOneMFA/PingOneMFATests/MockPingOneMFA.swift
```diff
@@ -0,0 +1,151 @@
...
+
+    static func getOneTimePasscode() async throws -> OtpCodeInfo {
+        getOneTimePasscodeCalled = true
+        if shouldThrowError {
+            throw PingOneMFAError(errorMessage)
+        }
+        return otpReturnValue
+    }
+
+    static func processPushNotification(userInfo: [AnyHashable: Any]) async throws -> PushNotification {
```
george-bafaloukas-forgerock: 🟡 **Mock method names drift from the real API:** `processPushNotification` / `getMobilePayload` / `processNotificationAction` here vs `processRemoteNotification` / `generateMobilePayload` / `processRemoteNotificationAction` in production. Since the mock isn't a protocol conformance, nothing keeps them in sync. If the intent is to exercise the real contract, having both conform to a shared protocol would turn drift into a compile error.

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -140,29 +203,57 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
...
-                    
-                    // Navigate to Push Notifications view
-                    NotificationCenter.default.post(
-                        name: NSNotification.Name("NavigateToPushNotifications"),
-                        object: nil
-                    )
-                } else {
-                    print("Notification was not processed (may be unsupported type)")
+        // Route to PingOneMFA if the category matches one registered by PingOneMFA.
+        if pingOneMFACategoryIdentifiers.contains(categoryIdentifier) {
```
george-bafaloukas-forgerock, line 204: 🔴 **Background banner actions can be silently lost.** `completionHandler()` at the end of this method fires while the actual approve/deny work runs in this detached `Task`. Calling the handler tells iOS the work is done, so the system is free to suspend the app and kill the in-flight `Task` before the request reaches PingOne. The PR explicitly advertises this flow ("tapping either completes the request"). Could we move `completionHandler()` to run *after* the awaited work (and grab a background-task assertion if needed)? Same pattern applies to `willPresent`.

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -110,23 +153,43 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
...
-            do {
-                let client = try await getInitializedPushClient()
 
-                // Process the notification - PushClient automatically extracts APNs payload
-                if let pushNotification = try await client.processNotification(userInfo: userInfoCopy) {
-                    print("Processed foreground push notification - ID: \(pushNotification.id), MessageID: \(pushNotification.messageId)")
-                } else {
-                    print("Foreground notification was not processed (may be unsupported type)")
+        // Route to PingOneMFA if the category matches one registered by PingOneMFA.
+        if pingOneMFACategoryIdentifiers.contains(categoryIdentifier) {
```
george-bafaloukas-forgerock: 🟠 **Cold-launch routing race.** `pingOneMFACategoryIdentifiers` is populated in a fire-and-forget `Task` in `didFinishLaunchingWithOptions`, but this routing check depends on it already being filled. When a notification *tap* launches the app cold, this can run first and the set is still empty → a valid PingOneMFA action is routed to `PushClient` and dropped. Could routing avoid depending on async-initialized in-memory state — e.g. derive the categories synchronously, or fall back to inspecting the payload?

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -93,6 +123,19 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
...
             }
         }
+
+        // Register raw APNS token with PingOneMFA
+        Task {
+            do {
+                try await ensurePingOneMFAInitialized()
+                try await PingOneMFA.setDeviceToken(deviceToken)
+                print("PingOneMFA device token registered successfully")
+            } catch let error as NSError where error.domain == "AppDelegate" {
```
george-bafaloukas-forgerock: 🟠 **This `catch` branch is dead and the message is misleading.** Neither `ensurePingOneMFAInitialized()` nor `PingOneMFA.setDeviceToken()` throws an `NSError` with domain `"AppDelegate"` — that domain only comes from `getInitializedPushClient()`. This looks copy-pasted from the PushClient registration above; it can never run, and it promises a "Will retry when client is ready" that doesn't exist. Safe to remove.

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -30,6 +34,24 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
         // Register for remote notifications
         application.registerForRemoteNotifications()
 
+        // Register PingOneMFA notification categories so the system can deliver
+        // actionable banner notifications (approve / deny actions).
+        Task {
+            do {
+                try await ensurePingOneMFAInitialized()
+                let pingOneMFACategories = PingOneMFA.getNotificationCategories()
+                pingOneMFACategoryIdentifiers = Set(pingOneMFACategories.map { $0.identifier })
```
george-bafaloukas-forgerock: 🟡 Related to the routing race above — this is where `pingOneMFACategoryIdentifiers` gets populated, asynchronously and after an `await`. Anything that reads it (the two delegate methods) needs to tolerate it being empty until this completes.

SampleApps/PingExample/PingExample/PingOneMFAOtpViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+        countdownTimer = nil
+
+        isLoading = true
+        errorMessage = nil
+
+        do {
+            let info = try await PingOneMFA.getOneTimePasscode()
+            otpInfo = info
+            // Clamp to 1 second minimum to avoid immediate re-fire.
+            countdown = max(1, info.secondsRemaining)
```
george-bafaloukas-forgerock: 🟡 The `max(1, info.secondsRemaining)` clamp masks a negative `secondsRemaining` coming from the SDK (see my comment on `getOneTimePasscode()`), but it then schedules a 1-second sleep before refetching — so if the SDK keeps returning an expired/negative window this becomes a tight ~1 Hz fetch loop. Fixing the clamp at the SDK source (`max(0, ...)`) is the cleaner spot.

SampleApps/PingExample/PingExample.xcodeproj/project.pbxproj
```diff
@@ -62,6 +70,7 @@
 		A525E7BA2DD7647200B02C17 /* PingExternalIdPGoogle.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = A525E7B22DD7647200B02C17 /* PingExternalIdPGoogle.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		A53B6E5B2F9A659B00B279CE /* ReadOnlyTextView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A53B6E5A2F9A659B00B279CE /* ReadOnlyTextView.swift */; };
 		A55177322FA502D100E575CF /* BooleanCollectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A55177312FA502D100E575CF /* BooleanCollectorView.swift */; };
+		A55177322FA502D100E575CF /* BooleanCollectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A55177312FA502D100E575CF /* BooleanCollectorView.swift */; };
```
george-bafaloukas-forgerock, line 73: 🟡 **Duplicate project entries.** This PR adds `BooleanCollectorView.swift` twice — both the `PBXBuildFile` (lines 72–73) and the `PBXFileReference` (721–722) appear with identical UUIDs. A pbxproj is a UUID-keyed map, so duplicate keys produce a malformed project that Xcode "repairs" nondeterministically and that churns on every save. Worth removing the duplicate lines.

george-bafaloukas-forgerock (changes requested):
Thanks for this — it's a substantial, well-structured module and the doc comments across the public surface are genuinely excellent. 🙏 I've left inline comments below from a fairly pedantic pass focused on the new `PingOneMFA` SDK code (plus the README, tests, and the sample-app integration).

A few things I really liked: the `Geo` enum cleanly isolates `PingOneGeo` so consumers never import `PingOneSDK`, the same goes for not leaking `NSError`/`PingOneSDKError`, and `getDeviceInfo()`'s "neither data nor errors" branch correctly avoids hanging the continuation — exactly the defensive resume that prevents continuation-leak crashes. The `AccountParserTests` edge-case coverage is thorough too.

The handful of items I'd consider blocking before merge:
- **Racy idempotency** in `initialize(geo:)` that contradicts its documented guarantee.
- **Background banner actions can be lost** because `completionHandler()` fires before the async approve/deny completes.
- **README example for `getDeviceInfo()` doesn't compile** (wrong return type).
- **Two tests that don't test what they claim** — including a data race inside the "thread safety" test.

Everything else is lower-stakes polish. Really nice work overall — happy to talk through any of these. 🚀

PingOneMFA/PingOneMFATests/PingOneMFATests.swift
gioraping: fixed

PingOneMFA/README.md
gioraping: fixed

SampleApps/PingExample/PingExample/AppDelegate.swift
gioraping: don't see an issue here

PingOneMFA/PingOneMFA/PushType.swift
```diff
@@ -0,0 +1,25 @@
...
+import Foundation
+
+/// Describes the interaction model required by a PingOne MFA push notification.
+public enum PushType: Sendable {
+    /// A standard authentication request. The user approves or denies with a single tap.
+    case `default`
+
+    /// A silent test push sent by the server to verify the device's push registration.
+    /// No user action is required.
+    case dry
```
george-bafaloukas-forgerock, line 20: Is `dry` the internal naming for the silent push notifications in PingOne MFA? If not I would suggest we name this `silent` instead of `dry`. If the internal `enum` on the linked SDK already has this pattern then we can leave it

PingOneMFA/README.md
```diff
@@ -0,0 +1,384 @@
...
+└──────────────────────────┼──────────────────────-┘
+                           │
+              ┌────────────▼─────────┐
+              │   PingOne MFA SDK    │
+              └──────────────────────┘
+```
+
+---
+
+## Add Dependency
```
george-bafaloukas-forgerock, line 54: Are we only supporting SPM and not CocoaPods?

PingOneMFA/PingOneMFA/PingOneMFAError.swift
```diff
@@ -0,0 +1,75 @@
...
+    init(_ error: Error) {
+        if let existing = error as? PingOneMFAError {
+            self = existing
+            return
+        }
+        let nsError = error as NSError
+        let userInfoString = nsError.userInfo
+            .map { "\($0.key)=\($0.value)" }
+            .joined(separator: ", ")
+        self.message = "Code=\(nsError.code) \"\(nsError.localizedDescription)\" UserInfo={\(userInfoString)}"
```
george-bafaloukas-forgerock: Since errorDescription returns message, error.localizedDescription carries a verbatim server-controlled dictionary into logs/UI — and it's redundant because internalErrorsList[].userInfo already holds it structurally. 
Keep message as Code=… "localizedDescription" and let consumers opt into the structured field.

PingOneMFA/PingOneMFA/PingOneMFAInternalError.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Foundation
+
+/// Structured representation of a single PingOne SDK error.
+///
+/// Instances are produced from the native SDK error type and exposed through
+/// ``PingOneMFAError/internalErrorsList``.
+///
+/// - SeeAlso: ``PingOneMFAError``
+public struct PingOneMFAInternalError: Sendable {
```
george-bafaloukas-forgerock: Can we add `Equatable`

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -110,28 +139,54 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
...
+        // before the async work completes.
+        let bgTask = UIApplication.shared.beginBackgroundTask(withName: "willPresent-processing")
+
+        // Route to PingOneMFA if the category matches one registered by PingOneMFA.
+        if pingOneMFACategoryIdentifiers.contains(categoryIdentifier) {
+            Task {
+                defer { UIApplication.shared.endBackgroundTask(bgTask) }
+                do {
+                    try await ensurePingOneMFAInitialized()
+                    let pingOneMFANotification: MFAPushNotification = try await PingOneMFA.processRemoteNotification(userInfo: userInfoCopy)
```
george-bafaloukas-forgerock: Silent `.dry` push still pops an approve/deny sheet — AppDelegate posts ShowPingOneMFANotification for every processed notification without checking pushType, and the view only auto-dismisses on isCancelAuthentication. A .dry push (documented as "no user action required") shows a Deny-only sheet. Filter .dry before posting.

george-bafaloukas-forgerock (commented):
Few more comments. Please confirm the lack of Podspec and maybe add a line in the Changelog

PingOneMFA/README.md
gioraping: Our SDK PingOneSDK only support SPM, so it's problematic to support cocoapods

PingOneMFA/PingOneMFA/PushType.swift
gioraping: dry is the correct naming, we use it for test purpuse in admin console

PingOneMFA/PingOneMFA/PingOneMFAError.swift
gioraping: fixed

SampleApps/PingExample/PingExample/AppDelegate.swift
gioraping: added check fro dry and that the pingOneMFANotification is not nil

PingOneMFA/PingOneMFA/PushType.swift
george-bafaloukas-forgerock: Thanks, that settles the naming — keeping `.dry` is the right call then. 👍

One heads-up though: a later commit (`e108fc3`) edited the *docs* to say `silent` even though the case stayed `.dry`, so now they disagree:

- `PushType.swift:18` lost the word and reads `/// A  test push sent by the server…` (double space where "silent" was).
- `PushNotification.swift:29` now documents `` - `.silent` — no user interaction needed `` — but there is no `.silent` case, so anyone writing `case .silent:` against `pushType` won't compile.

Could we revert both doc lines back to `.dry` (and restore the "silent test push" wording on `PushType.swift:18`)? That keeps the docs matching the (correct) `.dry` case. Minor, just doc text.

PingOneMFA/PingOneMFA/PingOneMFAError.swift
george-bafaloukas-forgerock: Thanks — the message no longer dumps the `userInfo`, that's exactly it. 🙏

Just one small leftover from the fix: dropping `UserInfo={…}` from the string orphaned the local that built it, so `PingOneMFAError.swift:44-46` now computes `userInfoString` and nothing reads it:

```swift
let userInfoString = nsError.userInfo        // ← computed…
    .map { "\($0.key)=\($0.value)" }
    .joined(separator: ", ")
self.message = "Code=\(nsError.code) \(nsError.localizedDescription)"   // …but not used here anymore
```

Swift will flag it ("initialization of immutable value 'userInfoString' was never used"). Safe to just delete those three lines — the full `userInfo` is still preserved structurally in `internalErrorsList[].userInfo`, so nothing is lost.

PingOneMFA/PingOneMFA/PingOneMfaAccount.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///   - errors: Non-fatal diagnostic errors returned by the upstream SDK.
+    public init(accounts: [PingOneMfaAccount], errors: [PingOneMFAError]? = nil) {
+        self.accounts = accounts
+        self.errors = errors
+    }
+}
+
+/// Represents a registered PingOne MFA account on this device.
+public struct PingOneMfaAccount: Sendable, Equatable {
+    /// The cloud region where this account is registered (e.g. `"NA"`, `"EU"`).
```
george-bafaloukas-forgerock: 🟡 Carried over from the earlier round and not yet addressed: the `region` example values here are still `"NA"` / `"EU"`, but `AccountParser` uses the raw top-level JSON keys, which the tests assert are `"NorthAmerica"` / `"Europe"` (e.g. `AccountParserTests` line 45/88), and the README was already corrected to those. Could we update this doc to `e.g. "NorthAmerica", "Europe"` so the source comment matches the parser, the tests, and the README? Just the comment text.

PingOneMFA/PingOneMFA/PingOneMFA.swift
```diff
@@ -0,0 +1,319 @@
...
+    /// Processes an incoming APNS push notification for MFA authentication.
+    ///
+    /// Extracts the `title` and `message` from `userInfo["aps"]["alert"]` (handling
+    /// both the string-alert and dict-alert forms of the APNS payload).
+    ///
+    /// - Parameter userInfo: The raw `userInfo` dictionary from
+    ///   `application(_:didReceiveRemoteNotification:fetchCompletionHandler:)`.
+    /// - Returns: A `PushNotification` holding the upstream `NotificationObject` plus
+    ///   the parsed title and message. in case of verification push during pairing it will return nil since the SDK handles it internally and no UI is needed.
+    /// - Throws: `PingOneMFAError` if the SDK call fails or returns no notification object.
```
george-bafaloukas-forgerock: 🟡 Now that the return type is `PushNotification?` and the no-object branch `resume(returning: nil)` instead of throwing, this doc block is a bit out of sync:

- The `- Throws:` line still says it throws "or returns no notification object" — but that case now returns `nil`, it doesn't throw. Only the real-error branch throws.
- The `- Returns:` sentence reads a little broken too ("…and message. in case of verification push during pairing it will return nil…") and over-narrows the nil case (the code returns nil whenever the SDK yields neither an object nor an error, not only during pairing).

Suggested wording:
```
/// - Returns: A `PushNotification` when the SDK yields a notification object;
///   `nil` when the SDK reports neither an object nor an error (handled internally).
/// - Throws: `PingOneMFAError` if the underlying SDK call reports an error.
```
The optional-return contract itself is a sound change — consumers route on it, so worth getting the doc precise.

PingOneMFA/README.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+### Push Notifications — Foreground
+
+When your app is in the foreground, process the incoming notification and present the appropriate UI based on push type:
+
+```swift
+// In application(_:didReceiveRemoteNotification:fetchCompletionHandler:):
+do {
+    let push = try await PingOneMFA.processRemoteNotification(userInfo: userInfo)
```
george-bafaloukas-forgerock: 🟠 This foreground example no longer compiles after the return-type change. `processRemoteNotification(userInfo:)` now returns `PushNotification?`, so `let push = try await … ` then `switch push.pushType` needs an unwrap. Could we adjust to e.g.:
```swift
if let push = try await PingOneMFA.processRemoteNotification(userInfo: userInfo) {
    switch push.pushType { … }
}
```
Same class of issue as the `getDeviceInfo` snippet we fixed last round — the signature change reintroduced it here.

PingOneMFA/README.md
```diff
@@ -0,0 +1,384 @@
...
+### `PingOneMFA`
+
+| Function | Returns | Description |
+|---|---|---|
+| `initialize(geo:)` | `async throws` | Configure the PingOne SDK for the selected service region. Idempotent after first success. |
+| `setDeviceToken(_:)` | `async throws` | Register or refresh the APNS push token with PingOne. |
+| `pair(pairingKey:)` | `async throws` | Pair a new MFA account. |
+| `getDeviceInfo()` | `async throws -> PingOneMFADeviceInfo` | Return all paired accounts and any non-fatal diagnostic errors. |
+| `getOneTimePasscode()` | `async throws -> OtpCodeInfo` | Return the current TOTP code and its remaining validity window. |
+| `processRemoteNotification(userInfo:)` | `async throws -> PushNotification` | Convert an APNS `userInfo` payload to a typed `PushNotification`. |
```
george-bafaloukas-forgerock: 🟡 Companion to the snippet above: the API table still lists the return type as `async throws -> PushNotification`, but it's now `-> PushNotification?`. Worth updating the table to the optional so it matches the code (and the `processRemoteNotificationAction` row just below, which already shows the optional).

SampleApps/PingExample/PingExample/AppDelegate.swift
```diff
@@ -113,25 +142,47 @@ class AppDelegate: NSObject, UIApplicationDelegate, @preconcurrency UNUserNotifi
...
+                        NotificationCenter.default.post(
+                            name: NSNotification.Name("ShowPingOneMFANotification"),
+                            object: nil,
+                            userInfo: ["notification": pingOneMFANotification]
+                        )
+                    }
                 }
             } catch {
-                print("Failed to process foreground push notification: \(error.localizedDescription)")
+                // PingOne SDK throws when the notification isn't a PingOne MFA push — fall through to PushClient.
```
george-bafaloukas-forgerock, line 170: 🟠 The `.dry` + nil check added here is great. One routing concern with the restructure, though: `willPresent` now calls `processRemoteNotification` for every foreground push and only falls through to `PushClient` in the `catch`. But the wrapper returns `nil` (not throws) for the "no object, no error" case — so the fall-through only happens if the binary SDK *throws* for a non-MFA push. If it instead returns `(nil, nil)` for a plain PingPush notification, we hit the `if let` (false), no error is thrown, the `catch` never runs, and the PingPush notification is silently dropped in the foreground.

The comment here asserts "PingOne SDK throws when the notification isn't a PingOne MFA push" — is that contract guaranteed? `didReceive` still routes by category match, which doesn't depend on that assumption, so the two callbacks now behave differently. Could we either (a) handle a `nil` result with an explicit `else` that falls through to `PushClient`, or (b) restore the category check here for symmetry with `didReceive`? Sample-app only, so not blocking — just a real behavior gap vs. the previous version.

george-bafaloukas-forgerock (commented):
A few small follow-ups from the latest push-flow changes — all minor/doc-level, nothing blocking. The redaction, `Equatable`, and the `.dry` sheet check all look good; thanks for those. 🙏 Left the specifics inline. (Also re: the podspec — saw the note that PingOneSDK is SPM-only, so disregard my earlier CocoaPods-parity comment; that's a reasonable constraint.)

SampleApps/PingExample/PingExample/AppDelegate.swift
gioraping: Yes, I understood that the behaviour is not correct. Our SDK guarantee that  it will throw exception if the received push is not PIngOne push.
nil is returned only in certain scenario that we receive ping push during pairing (test push) in this case our SDK returns nil.

PingOneMFA/PingOneMFA/PingOneMfaAccount.swift
gioraping: fixed

PingOneMFA/PingOneMFA/PingOneMFA.swift
gioraping: fixed

PingOneMFA/README.md
gioraping: fixed

PingOneMFA/README.md
gioraping: fixed

SampleApps/PingExample/PingExample/AppDelegate.swift
george-bafaloukas-forgerock: Perfect, that fully answers it — thanks for the detail. 🙏 If the binary SDK guarantees a throw for any non-PingOne push and only returns `nil` for an MFA test-push during pairing, then the catch-based fall-through to `PushClient` is correct and there's no silent-drop. I'm happy with the behaviour.

Only soft suggestion (non-blocking): since the whole routing now leans on that throw-vs-nil contract, it'd be worth capturing it as a one-line doc comment on `processRemoteNotification` (something like "throws if the push is not a PingOne MFA push; returns nil for an internally-handled test push during pairing") so the next person touching this doesn't have to rediscover it. Resolving on my end.

PingOneMFA/PingOneMFA/PingOneMFAError.swift
george-bafaloukas-forgerock: Friendly heads-up — I think this one got marked resolved, but the latest `readme fixes` commit (`3aa73e1`) only touched the README + region/Returns docs, so the dead local is still in `PingOneMFAError.swift:44-46` (the `userInfoString` build with no reader). It'll still emit the "never used" warning. Tiny one — just delete those three lines whenever you next touch the file. No rush. 🙂

PingOneMFA/PingOneMFA/PushType.swift
george-bafaloukas-forgerock: Same gentle heads-up here: the `readme fixes` commit didn't include these two doc lines, so `PushType.swift:18` still reads "A  test push" (double space) and `PushNotification.swift:29` still documents the non-existent `.silent` case. Just the doc text — reverting both to `.dry` whenever convenient will line them up with the (correct) case name. Not blocking. 👍

PingOneMFA/PingOneMFA/PushNotification.swift
```diff
@@ -0,0 +1,102 @@
...
+    internal let notificationObject: NotificationObject
+    /// The notification title, if provided by the server.
+    public let title: String?
+    /// The notification body message, if provided by the server.
+    public let message: String?
+
+    /// The interaction model required by this notification.
+    ///
+    /// Use this to determine the UI flow:
+    /// - `.silent` — no user interaction needed; the authentication completes silently.
```
george-bafaloukas-forgerock: This should be `.dry` right?

george-bafaloukas-forgerock (approved):
Looks good to me, please see the 1 new open comment. I would not merge this one yet, unless it has been cleared by QA

PingOneMFA/PingOneMFA/PushNotification.swift
gioraping: yes, fixed

rodrigoareis (approved):
LGTM

Package.swift
```diff
@@ -48,6 +48,7 @@ let package = Package(
         .library(name: "PingOath", targets: ["PingOath"]),
         .library(name: "PingPush", targets: ["PingPush"]),
         .library(name: "PingAuthMigration", targets: ["PingAuthMigration"]),
+        .library(name: "PingOneMFA", targets: ["PingOneMFA"]),
```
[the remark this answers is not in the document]
gioraping: cocoa pods not supported

PingOneMFA/PingOneMFA/OtpCodeInfo.swift
```diff
@@ -0,0 +1,29 @@
...
+    public let secondsRemaining: Int
+
+    /// Creates an `OtpCodeInfo` with the given code and remaining validity.
+    ///
+    /// - Parameters:
+    ///   - code: The current OTP code string.
+    ///   - secondsRemaining: Seconds until the code expires, or zero if already expired.
+    public init(code: String, secondsRemaining: Int) {
+        self.code = code
+        self.secondsRemaining = secondsRemaining
```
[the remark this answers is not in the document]
gioraping: This is just proxy object, it shouldn't add any logic

SampleApps/PingExample/PingExample/PingOneMFAPayloadView.swift
```diff
@@ -0,0 +1,135 @@
...
+                        colors: [Color.gray, Color.gray],
+                        startPoint: .topLeading,
+                        endPoint: .bottomTrailing
+                      )
+            )
+            .clipShape(RoundedRectangle(cornerRadius: 12))
+        }
+        .disabled(viewModel.payload == nil)
+        .accessibilityIdentifier("copyPayloadButton")
+    }
```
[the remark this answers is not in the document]
gioraping: just testing app, not importent

SampleApps/PingExample/PingExample/ContentView.swift
```diff
@@ -224,7 +245,8 @@ struct ContentView: View {
     @State private var navigateToPushNotifications = false
     @State private var showNoConfigAlert = false
     @State private var noConfigTypeName = ""
-    
+    @State private var pingOneMFANotification: MFAPushNotification? = nil
+
```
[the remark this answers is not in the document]
gioraping: this is false. 
MFAPushNotification is a public typealias for PushNotification, defined in [PushNotification.swift:14](vscode-webview://0dfcbbq92uiov3be3m0ej738npodkopfi0eria78t5a2l6uo9knp/PingOneMFA/PingOneMFA/PushNotification.swift#L14) and exported by the PingOneMFA module, which [ContentView.swift](vscode-webview://0dfcbbq92uiov3be3m0ej738npodkopfi0eria78t5a2l6uo9knp/SampleApps/PingExample/PingExample/ContentView.swift) already imports

PingOneMFA/PingOneMFA/PingOneMFAError.swift
```diff
@@ -0,0 +1,75 @@
...
+/// }
+/// ```
+public struct PingOneMFAError: Error, LocalizedError, Sendable {
+    public let message: String
+
+    /// Structured list of individual SDK errors, or `nil` when the failure did not originate
+    /// from the native SDK (e.g. an unexpected exception).
+    public let internalErrorsList: [PingOneMFAInternalError]?
+
+    public var errorDescription: String? { message }
```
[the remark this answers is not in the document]
gioraping: fixed

PR #162, opened 2026-06-02
SDKS-4249 QA for new DaVinci collectors

spetrov:
# JIRA Ticket

[SDKS-4249](https://pingidentity.atlassian.net/browse/SDKS-4249) [QA] Support for new DaVinci collectors

# Description

New tests added covering the following collectors/features:
- `ReadOnlyTextCollector` (agreement text)
- `BooleanCollector` (required "I agree" checkbox)
- `PhoneNumberCollector` (object payload format, extension field, end-to-end submission)
- `LabelCollector` (multiple rich text links, target field for open-in-new-tab)
- `PollingCollector` (timeout, completion, challenge approval, QR code challenge)
- Password policy validation (max repeat, max length, field-level policy source)

### DIFF
--- a/Davinci/DavinciTests/integration tests/FormFieldValidationTests.swift
+++ b/Davinci/DavinciTests/integration tests/FormFieldValidationTests.swift
@@ -17,103 +17,114 @@ import XCTest
 
 class FormFieldValidationTests: DaVinciBaseTests, @unchecked Sendable {
     private var daVinci: DaVinci!
-    
+
     override func setUp() async throws {
         self.configFileName = "Config"
         try await super.setUp()
-        
-        self.config.acrValues = "210f6b876da11c836ffc1c5fb38f3938"
-        
+
         daVinci = DaVinci.createDaVinci { config in
             config.logger = LogManager.standard
             config.module(PingDavinci.OidcModule.config) { oidcValue in
-                oidcValue.clientId = self.config.clientId
-                oidcValue.scopes = Set(self.config.scopes)
-                oidcValue.redirectUri = self.config.redirectUri
-                oidcValue.acrValues = self.config.acrValues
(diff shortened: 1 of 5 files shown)


Davinci/DavinciTests/integration tests/FormFieldValidationTests.swift
```diff
@@ -17,103 +17,114 @@ import XCTest
...
+
         daVinci = DaVinci.createDaVinci { config in
             config.logger = LogManager.standard
             config.module(PingDavinci.OidcModule.config) { oidcValue in
-                oidcValue.clientId = self.config.clientId
-                oidcValue.scopes = Set(self.config.scopes)
-                oidcValue.redirectUri = self.config.redirectUri
-                oidcValue.acrValues = self.config.acrValues
-                oidcValue.discoveryEndpoint = self.config.discoveryEndpoint
+                oidcValue.clientId = "a6859a12-5e6e-4f64-96bb-cc8577706bee"
```
george-bafaloukas-forgerock: Are these real values? If so maybe we should not hardcode

george-bafaloukas-forgerock (approved):
Minor comment for the env values to check

Davinci/DavinciTests/integration tests/FormFieldValidationTests.swift
spetrov: I updated the tests to load the configuration from a configuration file ([ConfigNew.json](https://github.com/ForgeRock/ping-ios-sdk/blob/SDKS-4249/Davinci/DavinciTests/Configuration/ConfigNew.json)). 
This configuration files need to be sanitized, and the pipeline should load the actual values from a secret store when triggered. We have a separate ticket for this task...

PR #163, opened 2026-06-04
Add unified JSON configuration support to iOS SDK

vahancouver:
[SDKS-5066](https://pingidentity.atlassian.net/browse/SDKS-5066)[iOS] Standardize SDK configuration

  Introduce a platform-neutral JSON schema for configuring DaVinci,
  Journey, OidcWebClient, and OidcDeviceClient from a single dictionary,
  enabling config-file-driven SDK initialisation across platforms.

  - Add JsonConfigParser / JsonConfigError / JsonConfigKey (PingOidc)
  - Add OidcClientConfig.apply(json:) for validated OIDC sub-dict parsing
  - Add createXxx(json:) factory on DaVinci, Journey, OidcWebClient,
    OidcDeviceClient — all return Result<T, Error>
  - Add LogManager.logger(forLevel:) mapping JSON log-level strings
  - Add JsonConfigLoader + JsonConfigPreviewView to sample app; load
    bundled JSON files from Configs/ folder reference at runtime
  - Cache parsed JSON dictionaries to avoid repeated disk reads
  - Gitignore Configs/ so local JSON files are never committed

### DIFF
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -15,6 +15,7 @@ import PingOidc
 import PingDavinciPlugin
 import PingCommons
 import PingNetwork
+import PingLogger
 
 public typealias DaVinci = Workflow
 public typealias DaVinciConfig = WorkflowConfig
@@ -120,5 +121,40 @@ extension DaVinci {
         block(config)
         
         return DaVinci(config: config)
+    }
+
+    /// Creates a DaVinci instance from a JSON dictionary.
+    ///
+    /// This factory parses and validates a platform-neutral JSON configuration and
+    /// delegates to `createDaVinci(block:)` with the extracted values. Unknown fields
+    /// (including `serverUrl`, `realm`, and `cookie`, which are Journey-specific) are
+    /// silently ignored for forward compatibility.
+    ///
+    /// - Parameter json: A `[String: Any]` dictionary conforming to the unified SDK
+    ///   configuration schema (see design doc for field reference).
+    /// - Returns: `.success(DaVinci)` on valid input, `.failure(JsonConfigError)` if a
+    ///   required field is absent or a field has the wrong type.
+    public static func createDaVinci(json: [String: Any]) -> Result<DaVinci, Error> {
+        do {
+            let p = JsonConfigParser(json)
+            let timeout = try p.timeoutSeconds()
+            let logger  = try p.logLevel()
+            let oidcDict: [String: Any] = try p.required(JsonConfigKey.oidc, field: JsonConfigKey.oidc)
+
+            let oidcConfig = OidcClientConfig()
+            oidcConfig.logger = logger
+            try oidcConfig.apply(json: oidcDict)
+
+            let daVinci = DaVinci.createDaVinci { daVinciConfig in
+                daVinciConfig.timeout = timeout
+                daVinciConfig.logger = logger
+                daVinciConfig.module(OidcModule.config) { moduleOidcConfig in
+                    moduleOidcConfig.update(with: oidcConfig)
+                }
+            }
+            return .success(daVinci)
+        } catch {
+            return .failure(error)
+        }
     }
 }
--- a/Davinci/Davinci/DaVinci.swift
+++ b/Davinci/Davinci/DaVinci.swift
@@ -138,12 +138,10 @@ extension DaVinci {
         do {
             let p = JsonConfigParser(json)
             let timeout = try p.timeoutSeconds()
-            let logger  = try p.logLevel()
+            let logger  = p.logLevel()
             let oidcDict: [String: Any] = try p.required(JsonConfigKey.oidc, field: JsonConfigKey.oidc)
 
-            let oidcConfig = OidcClientConfig()
-            oidcConfig.logger = logger
-            try oidcConfig.apply(json: oidcDict)
+            let oidcConfig = try OidcClientConfig.from(oidcJson: oidcDict, logger: logger)
 
             let daVinci = DaVinci.createDaVinci { daVinciConfig in
                 daVinciConfig.timeout = timeout
--- a/Journey/Journey/Journey.swift
+++ b/Journey/Journey/Journey.swift
@@ -194,6 +194,46 @@ public extension Journey {
         return await signOff()
     }
     
+    /// Creates a Journey instance from a JSON dictionary.
+    ///
+    /// This factory parses and validates a platform-neutral JSON configuration and
+    /// delegates to `createJourney(block:)` with the extracted values. Unknown fields
+    /// are silently ignored for forward compatibility.
+    ///
+    /// - Parameter json: A `[String: Any]` dictionary conforming to the unified SDK
+    ///   configuration schema (see design doc for field reference).
+    /// - Returns: `.success(Journey)` on valid input, `.failure(JsonConfigError)` if a
+    ///   required field is absent or a field has the wrong type.
+    static func createJourney(json: [String: Any]) -> Result<Journey, Error> {
+        do {
+            let p = JsonConfigParser(json)
+            let serverUrl: String = try p.required(JsonConfigKey.serverUrl, field: JsonConfigKey.serverUrl)
+            let realm      = try p.optional(JsonConfigKey.realm,      field: JsonConfigKey.realm,      default: JourneyConstants.realm)
+            let cookieName = try p.optional(JsonConfigKey.cookieName, field: JsonConfigKey.cookieName, default: JourneyConstants.cookie)
+            let timeout    = try p.timeoutSeconds()
+            let logger     = try p.logLevel()
+            let oidcDict: [String: Any] = try p.required(JsonConfigKey.oidc, field: JsonConfigKey.oidc)
+
+            let oidcConfig = OidcClientConfig()
+            oidcConfig.logger = logger
+            try oidcConfig.apply(json: oidcDict)
+
+            let journey = Journey.createJourney { journeyConfig in
+                journeyConfig.serverUrl = serverUrl
+                journeyConfig.realm = realm
+                journeyConfig.cookie = cookieName
+                journeyConfig.timeout = timeout
+                journeyConfig.logger = logger
+                journeyConfig.module(OidcModule.config) { moduleOidcConfig in
+                    moduleOidcConfig.update(with: oidcConfig)
+                }
+            }
+            return .success(journey)
+        } catch {
+            return .failure(error)
+        }
+    }
+
     /// Sends a request using the configured HTTP client and wraps the response.
     ///
     /// - Parameter request: The request to send.
--- a/Journey/Journey/Journey.swift
+++ b/Journey/Journey/Journey.swift
@@ -213,16 +213,12 @@ public extension Journey {
             let realm      = try jp.optional(JsonConfigKey.realm,      field: "\(JsonConfigKey.journey).\(JsonConfigKey.realm)",      default: JourneyConstants.realm)
             let cookieName = try jp.optional(JsonConfigKey.cookieName, field: "\(JsonConfigKey.journey).\(JsonConfigKey.cookieName)", default: JourneyConstants.cookie)
             let timeout    = try p.timeoutSeconds()
-            let logger     = try p.logLevel()
+            let logger     = p.logLevel()
             let oidcDict: [String: Any]? = try p.optionalValue(JsonConfigKey.oidc, field: JsonConfigKey.oidc)
 
-            let oidcConfig: OidcClientConfig? = try {
-                guard let oidcDict else { return nil }
-                let config = OidcClientConfig()
-                config.logger = logger
-                try config.apply(json: oidcDict)
-                return config
-            }()
+            let oidcConfig: OidcClientConfig? = try oidcDict.map {
+                try OidcClientConfig.from(oidcJson: $0, logger: logger)
+            }
 
             let journey = Journey.createJourney { journeyConfig in
                 journeyConfig.serverUrl = serverUrl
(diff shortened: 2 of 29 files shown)


Oidc/Oidc/JsonConfig.swift
```diff
@@ -0,0 +1,131 @@
...
+    // MARK: - oidc sub-dict — required
+    public static let clientId = "clientId"
+    public static let discoveryEndpoint = "discoveryEndpoint"
+    public static let redirectUri = "redirectUri"
+    public static let scopes = "scopes"
+
+    // MARK: - oidc sub-dict — optional
+    public static let signOutRedirectUri = "signOutRedirectUri"
+    public static let refreshThreshold = "refreshThreshold"
+    public static let par = "par"
```
vibhorgoswami, line 36: What is this value? Is this part of iOS only?

Davinci/Davinci/DaVinci.swift
```diff
@@ -121,4 +122,39 @@ extension DaVinci {
...
+    /// This factory parses and validates a platform-neutral JSON configuration and
+    /// delegates to `createDaVinci(block:)` with the extracted values. Unknown fields
+    /// (including `serverUrl`, `realm`, and `cookie`, which are Journey-specific) are
+    /// silently ignored for forward compatibility.
+    ///
+    /// - Parameter json: A `[String: Any]` dictionary conforming to the unified SDK
+    ///   configuration schema (see design doc for field reference).
+    /// - Returns: `.success(DaVinci)` on valid input, `.failure(JsonConfigError)` if a
+    ///   required field is absent or a field has the wrong type.
+    public static func createDaVinci(json: [String: Any]) -> Result<DaVinci, Error> {
```
vibhorgoswami, line 137: Question: Are we allowing the user to add custom config using the `DaVinciConfig`?

Oidc/Oidc/JsonConfig.swift
vahancouver: it's for enabling PAR. shouldn't be iOS specific

Davinci/Davinci/DaVinci.swift
vahancouver: no, DaVinciConfig is not used here. just what's in the JSON

Oidc/Oidc/OpenIdConfiguration.swift
```diff
@@ -14,19 +14,19 @@ import Foundation
 /// Struct representing the OpenID Connect configuration.
 public struct OpenIdConfiguration: Codable, Sendable {
     /// The URL of the authorization endpoint.
-    public let authorizationEndpoint: String
+    public var authorizationEndpoint: String
     /// The URL of the token endpoint.
-    public let tokenEndpoint: String
+    public var tokenEndpoint: String
```
george-bafaloukas-forgerock, line 19: The switch from `let` to `var` is necessary for the `openIdOverride` inout closure, but it opens a second mutation path that didn't exist before: `OidcClientConfig.openId` is `public var`, so any external caller can now write directly to discovered endpoints after `oidcInitialize()` completes (e.g. `config.openId?.tokenEndpoint = "…"`). The class is `@unchecked Sendable`, so there's no compiler safety net here.

Would it make sense to keep `openId` as `public private(set) var`? That way the value is still readable from outside, the inout closure can still mutate it from within the module, but direct external writes are blocked and the `openIdOverride` closure stays the single intended mutation point.

Oidc/Oidc/OidcClientConfig.swift
```diff
@@ -133,7 +133,7 @@ public class OidcClientConfig: @unchecked Sendable {
     
     /// Merges another configuration into this one.
     /// - Parameter other: The other configuration to merge.
-    func update(with other: OidcClientConfig) {
+    public func update(with other: OidcClientConfig) {
         self.openId = other.openId
```
george-bafaloukas-forgerock, line 169: Making `update(with:)` public is needed for the cross-module factory methods, but it's quite a wide door to open: the method replaces *every* field on the config including `storage` (the keychain reference) and `openId`. If a caller invokes this on a config that's already been passed to a running `OidcWebClient` or `DaVinci` workflow, they get a hard-to-diagnose situation where in-flight token reads suddenly start hitting a different (empty) keychain store.

The class is `@unchecked Sendable`, so Swift 6 won't flag a concurrent call either. Would scoping this more narrowly work? For example, leaving `update(with:)` internal and instead giving the factory methods a `@_spi` or package-access variant — or exposing a more limited "apply config snapshot" API that only copies the JSON-derived fields. Just worth thinking about before this lands as a public API surface.

Oidc/Oidc/OidcClientConfig.swift
```diff
@@ -155,4 +155,111 @@ public class OidcClientConfig: @unchecked Sendable {
...
+
+        // All validation passed — apply to self
+        self.clientId = clientId
+        self.discoveryEndpoint = discoveryEndpoint
+        self.scopes = parsedScopes
+        self.redirectUri = redirectUri
+        self.refreshThreshold = Int64(refreshThresholdInt)
+        self.par = parsedPar
+        self.loginHint = loginHint
+        self.state = state
```
george-bafaloukas-forgerock, line 269: This is documented in the `apply(json:)` doc comment — nice that it's called out — but the consequence is sharper than it first appears. If a consumer builds a device client via the builder API and sets a custom `openIdOverride` (e.g. to inject `deviceAuthorizationEndpoint`), and then a subsequent configuration-reload calls `apply(json:)` with a JSON that happens to include *any* `openId` key, their closure is silently dropped and the device auth endpoint reverts to whatever discovery returned.

A merge approach would be safer here: capture the existing `openIdOverride`, run it first in the new closure, then apply the JSON-derived overrides on top. That way caller-supplied overrides aren't clobbered unless the JSON explicitly covers the same endpoint key.

Oidc/Oidc/JsonConfig.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+        return value
+    }
+
+    /// Parses `"log"` as a log-level string and returns the matching `Logger`. Defaults to `LogManager.none` when absent.
+    public func logLevel() throws -> Logger {
+        guard let raw = json[JsonConfigKey.log] else { return LogManager.logger }
+        guard let level = raw as? String else {
+            throw JsonConfigError.invalidType(field: JsonConfigKey.log, expected: "string")
+        }
```
george-bafaloukas-forgerock: This is a hard failure for a purely advisory field. If a consumer passes `"log": 1` (easy to do when generating configs programmatically), every factory method returns `.failure` and the SDK can't be initialised at all — even though the log level has zero effect on authentication correctness.

Would it be reasonable to treat a wrong-type `"log"` value as a soft fault and fall back to `NoneLogger` instead? Something like:
```swift
guard let level = raw as? String else {
    return LogManager.none // wrong type — ignore rather than blocking init
}
```
The field isn't security-relevant, so silently ignoring a bad value feels better than blocking the whole SDK.

Oidc/Oidc/OidcClientConfig.swift
```diff
@@ -155,4 +155,111 @@ public class OidcClientConfig: @unchecked Sendable {
...
+        let uiLocales:  String? = try p.optionalValue(JsonConfigKey.uiLocales,  field: f(JsonConfigKey.uiLocales))
+        let acrValues:  String? = try p.optionalValue(JsonConfigKey.acrValues,  field: f(JsonConfigKey.acrValues))
+
+        // --- additionalParameters ---
+        var parsedAdditional = [String: String]()
+        if let rawDict: [String: Any] = try p.optionalValue(JsonConfigKey.additionalParameters, field: f(JsonConfigKey.additionalParameters)) {
+            for (key, value) in rawDict {
+                guard let stringValue = value as? String else {
+                    throw JsonConfigError.invalidType(field: f("\(JsonConfigKey.additionalParameters).\(key)"), expected: "string")
+                }
```
george-bafaloukas-forgerock, line 241: 💡 Minor suggestion: `endpointSetters` holds 8 closures that capture no instance state — they only reference `$0` and `$1`. Defining it as a local `let` inside `apply(json:)` means a new array + 8 closure objects are allocated on every call, and the resulting `openIdOverride` closure keeps that array alive for the config's lifetime.

Moving it to a `private static let` on `OidcClientConfig` (or a file-level private constant) would make it a one-time allocation with no functional change.

Davinci/Davinci/DaVinci.swift
```diff
@@ -121,4 +122,39 @@ extension DaVinci {
...
+        do {
```
george-bafaloukas-forgerock, line 138: 💡 Minor suggestion: this exact three-line block appears identically in all four factory methods (`createDaVinci`, `createJourney`, `createOidcWebClient`, `createOidcDeviceClient`):
```swift
let oidcConfig = OidcClientConfig()
oidc Config.logger = logger
try oidcConfig.apply(json: oidcDict)
```
If a new field is ever added to the JSON-to-config handshake, all four files need updating — easy to miss one. A small static helper on `OidcClientConfig` (e.g. `static func fromJson(_ json: [String: Any], logger: Logger) throws -> OidcClientConfig`) would keep this in one place and make the factory methods a bit easier to read too.

witrisna (approved):

Oidc/Oidc/OpenIdConfiguration.swift
vahancouver: Changed to public private(set) var — external reads still work, but direct writes are 
  blocked. Added an internal setOpenId(_:) method as a test-only injection point (replaces the 
 direct assignments in 4 test files). openIdOverride is now the sole public mutation path.

Oidc/Oidc/OidcClientConfig.swift
vahancouver: Can't make it internal — callers are in PingDavinci, PingJourney, etc. (separate modules). Added a doc comment instead, explicitly calling out that this is a module-wiring primitive and warning  against calling it on a live config.
 
 If you think the doc comment is not enough and we should use a different more limited public method, I can look into it more.

Oidc/Oidc/OidcClientConfig.swift
vahancouver: Fixed — the existing openIdOverride is now captured and run first, then JSON-derived overrides are applied on top. JSON values win for any key they cover. Updated the doc comment to describe merge semantics.

Oidc/Oidc/JsonConfig.swift
vahancouver: Fixed — logLevel() is no longer throws. A wrong-type "log" value now silently falls back to  LogManager.none rather than blocking SDK init. Removed try from all four call sites, and updated  the four failure_logWrongType tests to assert .success instead.

Oidc/Oidc/OidcClientConfig.swift
vahancouver: Done — moved to private static let (with nonisolated(unsafe) for Swift 6 concurrency). One allocation at class load time, no functional change.

Davinci/Davinci/DaVinci.swift
vahancouver: Extracted into public static func from(oidcJson:logger:) throws -> OidcClientConfig. All four  factories now call this helper.

george-bafaloukas-forgerock (approved):
LGTM

Journey/Journey/Journey.swift
```diff
@@ -194,6 +194,50 @@ public extension Journey {
...
+    ///
+    /// This factory parses and validates a platform-neutral JSON configuration and
+    /// delegates to `createJourney(block:)` with the extracted values. Unknown fields
+    /// are silently ignored for forward compatibility.
+    ///
+    /// - Parameter json: A `[String: Any]` dictionary conforming to the unified SDK
+    ///   configuration schema (see design doc for field reference).
+    /// - Returns: `.success(Journey)` on valid input, `.failure(JsonConfigError)` if a
+    ///   required field is absent or a field has the wrong type.
+    static func createJourney(json: [String: Any]) -> Result<Journey, Error> {
```
spetrov, line 207: Should this be `public`?

Oidc/Oidc/OidcWebClient.swift
```diff
@@ -50,6 +51,38 @@ public extension OidcWebClient {
...
+    ///
+    /// This factory parses and validates a platform-neutral JSON configuration and
+    /// delegates to `createOidcWebClient(block:)` with the extracted values. Unknown
+    /// fields are silently ignored for forward compatibility.
+    ///
+    /// - Parameter json: A `[String: Any]` dictionary conforming to the unified SDK
+    ///   configuration schema (see design doc for field reference).
+    /// - Returns: `.success(OidcWebClient)` on valid input, `.failure(JsonConfigError)` if a
+    ///   required field is absent or a field has the wrong type.
+    static func createOidcWebClient(json: [String: Any]) -> Result<OidcWebClient, Error> {
```
spetrov, line 64: Should this be `public`?

spetrov (approved):
LGTM, but please check my questions regarding the access modifier of `createJourney(json:)` and `createOidcWebClient(json:)`...

Journey/Journey/Journey.swift
vahancouver: any method in a public extension is public by default. Xcode gives warning when I explicitly set it public "'public' modifier is redundant for static method declared in a public extension"

Oidc/Oidc/OidcWebClient.swift
vahancouver: any method in a public extension is public by default. Xcode gives warning when I explicitly set it public "'public' modifier is redundant for static method declared in a public extension"

Journey/Journey/Journey.swift
spetrov: got it! thanks!

PR #165, opened 2026-06-11
SDKS-5034 Add AGENTS.md as canonical AI agent context

rodrigoareis:
# JIRA Ticket

[SDKS-5034](https://pingidentity.atlassian.net/browse/SDKS-5034) "[iOS] Add context to the SDK repository"

# Description

Add the open-source, tool-agnostic `AGENTS.md` standard at the repo root and ensure `CLAUDE.md` with a one-line redirect file** (`See AGENTS.md.`) rather than a symlink. This avoids symlink portability issues (Windows `core.symlinks=false`) while keeping maintenance burden at zero — the redirect only needs updating if `AGENTS.md` is renamed. Claude Code reads both `AGENTS.md` and `CLAUDE.md` natively, so it picks up the full content from `AGENTS.md` directly.

# Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -111,7 +111,7 @@ These lightweight protocol layers (`Collector`, `ContinueNode`, `Submittable`, `
 ### Dependency Hierarchy
 
 ```
-PingLogger, PingStorage          ← foundation (no deps)
+PingLogger, PingStorage ← foundation (no deps)
 PingNetwork ← PingLogger
 PingCommons ← PingLogger
 PingBrowser ← PingLogger
@@ -129,6 +129,7 @@ PingBinding ← PingCommons, PingDeviceId, PingJourneyPlugin
 PingOath ← PingTamperDetector
 PingPush ← PingNetwork, PingTamperDetector
 PingAuthMigration ← PingOath, PingPush
+PingDeviceClient ← PingCommons
 ```
 
 ### Storage Layer
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -129,7 +129,7 @@ PingBinding ← PingCommons, PingDeviceId, PingJourneyPlugin
 PingOath ← PingTamperDetector
 PingPush ← PingNetwork, PingTamperDetector
 PingAuthMigration ← PingOath, PingPush
-PingDeviceClient ← PingCommons
+PingDeviceClient ← PingCommons, PingNetwork
 ```
 
 ### Storage Layer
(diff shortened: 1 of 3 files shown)


george-bafaloukas-forgerock (approved):
LGTM

CLAUDE.md
```diff
@@ -0,0 +1 @@
+See AGENTS.md.
```
vahancouver: you can double check, but apparently this wouldn't work. you need to use the file reference like this
`@AGENTS.md`

AGENTS.md
```diff
@@ -0,0 +1,153 @@
...
+PingReCaptchaEnterprise ← PingCommons, PingJourneyPlugin, ReCaptcha Enterprise (iOS only)
+PingExternalIdP ← PingBrowser, PingDavinciPlugin, PingJourneyPlugin
+PingExternalIdP{Apple,Google,Facebook} ← PingExternalIdP (+ vendor SDK on iOS only)
+PingProtect ← PingDavinciPlugin, PingJourneyPlugin, PingOneSignals (iOS only)
+PingFido ← PingLogger, PingCommons, PingDavinciPlugin, PingJourneyPlugin
+PingBinding ← PingCommons, PingDeviceId, PingJourneyPlugin
+PingOath ← PingTamperDetector
+PingPush ← PingNetwork, PingTamperDetector
+PingAuthMigration ← PingOath, PingPush
+```
```
vahancouver, line 133: PingDeviceClient ←PingCommons , PingNetwork is missing

AGENTS.md
```diff
@@ -0,0 +1,153 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents (Claude Code, Cursor, GitHub Copilot, Gemini, etc.) when working with code in this repository. It follows the open AGENTS.md convention.
```
vahancouver, line 3: Should we mention Gemini here if we are still ignoring GEMINI.md file?

vahancouver (changes requested):
left some comments

AGENTS.md
rodrigoareis: PingNetwork is on row #115

AGENTS.md
rodrigoareis: I think it's ok to mention it in general, even if we are not including a `GEMINI.md` file

AGENTS.md
vahancouver: Sorry I think I misspoke. I meant `PingDeviceClient` also depends on `PingNetwork`. that part is missing

vahancouver (approved):
LGTM

PR #166, opened 2026-06-16
Sdks 5021 par integration tests

vahancouver:
[SDKS-5021](https://pingidentity.atlassian.net/browse/SDKS-5021) [QA][Android][iOS] PAR

Adds real-server E2E test coverage for Pushed Authorization Requests across all three OIDC entry points on iOS: Journey, DaVinci, and OidcWebClient.

### DIFF
--- a/Davinci/Davinci.xcodeproj/project.pbxproj
+++ b/Davinci/Davinci.xcodeproj/project.pbxproj
@@ -68,6 +68,8 @@
 		AA4785031F0000000000AA01 /* DaVinciDeviceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA4785021F0000000000AA01 /* DaVinciDeviceTests.swift */; };
 		BB4928012F0900000000BB01 /* ReadOnlyTextCollector.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB4928002F0900000000BB01 /* ReadOnlyTextCollector.swift */; };
 		BB4928032F0900000000BB01 /* ReadOnlyTextCollectorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB4928022F0900000000BB01 /* ReadOnlyTextCollectorTests.swift */; };
+		BBBBBBBB00000001FAA0E2E0 /* DaVinciPARE2ETest.swift in Sources */ = {isa = PBXBuildFile; fileRef = BBBBBBBB00000000FAA0E2E0 /* DaVinciPARE2ETest.swift */; };
+		BBBBBBBB00000003FAA0E2E0 /* RecordingHttpClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = BBBBBBBB00000002FAA0E2E0 /* RecordingHttpClient.swift */; };
 		EC5C81C92ECF7A1700C91570 /* PingDavinciPlugin.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EC5C81C82ECF7A1700C91570 /* PingDavinciPlugin.framework */; };
 		EC5C81CA2ECF7A1700C91570 /* PingDavinciPlugin.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = EC5C81C82ECF7A1700C91570 /* PingDavinciPlugin.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		EC5C81FC2ED0CAA100C91570 /* Collectors.swift in Sources */ = {isa = PBXBuildFile; fileRef = EC5C81FB2ED0CAA100C91570 /* Collectors.swift */; };
@@ -170,6 +172,8 @@
 		AA4785021F0000000000AA01 /* DaVinciDeviceTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DaVinciDeviceTests.swift; sourceTree = "<group>"; };
 		BB4928002F0900000000BB01 /* ReadOnlyTextCollector.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadOnlyTextCollector.swift; sourceTree = "<group>"; };
 		BB4928022F0900000000BB01 /* ReadOnlyTextCollectorTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadOnlyTextCollectorTests.swift; sourceTree = "<group>"; };
+		BBBBBBBB00000000FAA0E2E0 /* DaVinciPARE2ETest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DaVinciPARE2ETest.swift; sourceTree = "<group>"; };
+		BBBBBBBB00000002FAA0E2E0 /* RecordingHttpClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = RecordingHttpClient.swift; path = ../Network/NetworkTests/RecordingHttpClient.swift; sourceTree = SOURCE_ROOT; };
 		EC5C81C82ECF7A1700C91570 /* PingDavinciPlugin.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = PingDavinciPlugin.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		EC5C81FB2ED0CAA100C91570 /* Collectors.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Collectors.swift; sourceTree = "<group>"; };
 		EC7279AD2DF98CD6005D9E28 /* Config.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Config.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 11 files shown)


Oidc/OidcTests/IntegrationTests/OidcWebClientE2ETests.swift
```diff
@@ -0,0 +1,207 @@
...
+        try await super.tearDown()
+    }
+
+    // MARK: - AIC (Journey server)
+
+    func testOidcWebClientAICWithoutPAR() async throws {
+        let web = makeWebClient(config: aicConfig, additionalParameters: ["foo": "bar"], par: false)
+
+        _ = try? await web.authorize()
+
```
george-bafaloukas-forgerock, line 56: **[Review] `try?` masks pre-browser errors, producing misleading failures**

`try?` discards any error thrown before `CapturingBrowser.launch()` is called. If the live discovery endpoint is unreachable, credentials in the config JSON have expired, or the PAR POST fails, `authorize()` throws before the browser step and `browser.launchedURL` is never set. The subsequent `XCTUnwrap` then fails with "Browser was not launched" — with no trace of the actual cause.

The same pattern repeats on all four `_ = try? await web.authorize()` call sites in this file (~lines 55, 74, 105, 126).

**Suggested fix:** propagate the real error rather than swallowing it:
```swift
// Option A — let the real error propagate (preferred if authorize() is expected to fail at
// the redirect step, which XCTest will surface as a known-failure rather than crash):
do {
    _ = try await web.authorize()
} catch {
    // expected — CapturingBrowser doesn't complete the redirect
}

// Option B — keep try? but assert the error type you expect to be swallowed:
let err = await XCTAssertThrowsError(try await web.authorize())
// ... only proceed if error is the expected BrowserCancelledError
```

Network/NetworkTests/RecordingHttpClient.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// Parses a `application/x-www-form-urlencoded` body into a key→value dictionary.
+    /// Returns an empty dictionary for any other content type.
+    func formFields() -> [String: String] {
+        guard let body, let str = String(data: body, encoding: .utf8) else { return [:] }
+        var out: [String: String] = [:]
+        for pair in str.split(separator: "&") {
+            let kv = pair.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+            guard kv.count == 2 else { continue }
+            let k = String(kv[0]).removingPercentEncoding ?? String(kv[0])
+            let v = String(kv[1]).removingPercentEncoding ?? String(kv[1])
```
george-bafaloukas-forgerock: **[Review] `formFields()` silently misreads PAR body values — `+` is not decoded as space**

RFC 1866 / HTML form encoding uses `+` to represent spaces in `application/x-www-form-urlencoded` bodies. `removingPercentEncoding` only handles `%XX` sequences and leaves `+` as a literal plus character.

**Concrete scenario:** If the SDK encodes the PAR push body with `scope=openid+email+address`, `formFields()` returns `{"scope": "openid+email+address"}` instead of `{"openid email address"}`. The key-presence assertions (`XCTAssertNotNil(parFields["scope"])`) still pass silently; any equality check on the value is wrong.

**Fix:**
```swift
let k = String(kv[0]).replacingOccurrences(of: "+", with: " ").removingPercentEncoding ?? String(kv[0])
let v = String(kv[1]).replacingOccurrences(of: "+", with: " ").removingPercentEncoding ?? String(kv[1])
```

Network/NetworkTests/RecordingHttpClient.swift
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    func request() -> HttpRequest {
+        inner.request()
+    }
+
+    func request(request: HttpRequest) async throws -> HttpResponse {
+        record(request)
+        return try await inner.request(request: request)
```
george-bafaloukas-forgerock, line 101: **[Review] Recording happens before interceptors run — snapshot may be incomplete**

`record(request)` is called here, before `inner.request(request: request)` dispatches the request. Inside `URLSessionHttpClient`, `requestInterceptors` mutate the `HttpRequest` object before `buildURLRequest()` is called for actual dispatch. Any headers or mutations injected by interceptors (auth tokens, signing, additional PKCE fields) will be absent from the captured `RecordedRequest`.

For the current tests this is harmless if the inner client has no interceptors registered. But if interceptors are ever added (e.g. a signing interceptor for the PAR POST), the recorded snapshot will diverge from what is actually sent over the wire, and PAR header assertions would silently test the wrong state.

**Consider** adding a comment here documenting that the snapshot is pre-interceptor, or hooking into `URLSessionDelegate` / a `URLProtocol` subclass to capture post-interceptor wire data if header fidelity becomes important.

Oidc/OidcTests/IntegrationTests/Config/OidcE2EConfig.swift
```diff
@@ -0,0 +1,49 @@
...
+    let discoveryEndpoint: String
+    let scopes: [String]
+    let redirectUri: String
+    let acrValues: String
+    /// Optional: e.g. "pi.flow" for PingOne DaVinci flows; empty for AIC.
+    let responseMode: String
+
+    init(_ fileName: String) throws {
+        guard let path = Bundle(for: OidcWebClientE2ETests.self)
+                .path(forResource: fileName, ofType: "json")
```
george-bafaloukas-forgerock, line 26: **[Review] `Bundle(for: OidcWebClientE2ETests.self)` hardcodes the wrong bundle for cross-target reuse**

The bundle lookup is correct today because `OidcE2EConfig` is only instantiated from within `OidcWebClientE2ETests` (same target). However, the struct looks general-purpose, and if it is ever instantiated from `DaVinciTests` or `JourneyTests` (which are separate bundles), `Bundle(for: OidcWebClientE2ETests.self)` will resolve to the OidcTests bundle — and the JSON resource files won't be found there. The init will throw `OidcE2EConfigError.fileNotFound` with no hint pointing to the wrong bundle.

**Options:**
```swift
// Option A — pass the bundle explicitly (most flexible):
init(_ fileName: String, bundle: Bundle = Bundle(for: OidcWebClientE2ETests.self)) throws { ... }

// Option B — keep it scoped; add a comment that it is OidcTests-only:
// NOTE: this struct is intentionally OidcTests-scoped. If you need config loading
// in another target, copy+adjust the Bundle reference.
```

Davinci/Davinci.xcodeproj/project.pbxproj
```diff
@@ -171,6 +173,8 @@
 		AA4785021F0000000000AA01 /* DaVinciDeviceTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DaVinciDeviceTests.swift; sourceTree = "<group>"; };
 		BB4928002F0900000000BB01 /* ReadOnlyTextCollector.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadOnlyTextCollector.swift; sourceTree = "<group>"; };
 		BB4928022F0900000000BB01 /* ReadOnlyTextCollectorTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadOnlyTextCollectorTests.swift; sourceTree = "<group>"; };
+		BBBBBBBB00000000FAA0E2E0 /* DaVinciPARE2ETest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DaVinciPARE2ETest.swift; sourceTree = "<group>"; };
+		BBBBBBBB00000002FAA0E2E0 /* RecordingHttpClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = RecordingHttpClient.swift; path = ../Network/NetworkTests/RecordingHttpClient.swift; sourceTree = SOURCE_ROOT; };
 		DA0000012F0000000000DA01 /* DeviceAuthorizationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DeviceAuthorizationTests.swift; sourceTree = "<group>"; };
```
george-bafaloukas-forgerock, line 184: **[Review] Cross-project relative path for `RecordingHttpClient.swift` will break silently on directory rename**

This entry (and the equivalent in `Journey/Journey.xcodeproj` and `Oidc/Oidc.xcodeproj`) references `RecordingHttpClient.swift` via:
```
path = ../Network/NetworkTests/RecordingHttpClient.swift; sourceTree = SOURCE_ROOT;
```

If the `Network/` directory is ever renamed (e.g. to `PingNetwork/` to match the module name), all three cross-project references break. Xcode does not flag this at edit time — it surfaces only as a "file not found" build error the next time DaVinci, Journey, or Oidc test schemes are built.

**Suggested alternatives:**
1. Move `RecordingHttpClient.swift` to a dedicated SPM test helper target shared by all three modules.
2. Copy it into each test target so each project owns its file.
3. If the current layout is intentional, add a comment in the pbxproj and update the repo's contribution notes to warn that renaming `Network/` requires updating three pbxproj files.

george-bafaloukas-forgerock (commented):
Code review findings from automated analysis. Five issues noted below as inline comments.

Oidc/OidcTests/IntegrationTests/OidcWebClientE2ETests.swift
vahancouver: Replaced with do { … } catch { } so actual errors (bad creds, network down, PAR failure) propagate through XCTest instead of silently swallowing and giving a misleading "Browser was not launched" failure.

Network/NetworkTests/RecordingHttpClient.swift
vahancouver: Added .replacingOccurrences(of: "+", with: " ") before removingPercentEncoding on both key and value, per RFC 1866.

Network/NetworkTests/RecordingHttpClient.swift
vahancouver: Added a comment documenting that the captured request is pre-interceptor so future maintainers aren't confused if a signing interceptor is ever added.

Oidc/OidcTests/IntegrationTests/Config/OidcE2EConfig.swift
vahancouver: Added a NOTE comment; the struct is only used in OidcTests today so no behavior change needed.

Davinci/Davinci.xcodeproj/project.pbxproj
vahancouver: No change; the file already has a comment about the cross-project pattern, and this is a known repo layout trade-off. 
I prefer this over having the same file in 3 places or adding anew module. If folder name changes, the test will not build and Xcode will complaint that it can't find the file- easy fix in that case

george-bafaloukas-forgerock (approved):
LGTM

spetrov (approved):
LGTM! 👍🏻

PR #167, opened 2026-06-17
[fix] SDKS-5172 Defensive recovery for EncryptorError after iCloud device migration

george-bafaloukas-forgerock:
# JIRA Ticket

[JIRA](https://bugster.forgerock.org/jira/browse/SDKS-5172) "[TRIAGE] OIDC authentication where the app returns: \"Authorization error: The operation couldn't be completed. (PingStorage.EncryptorError error 1)\""

# Description

After an iCloud backup / Quick Start device migration, the OIDC token ciphertext migrates to the new device but the Secure Enclave private key does not. This causes every call to `OidcClient.token()` to fail permanently with `EncryptorError.failedToDecrypt`. The fix has two complementary parts: (1) catch the decryption failure at the `token()` call site, delete the unreadable ciphertext, and fall through to re-authentication; (2) set `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` on `SecItemAdd` so that newly written tokens are device-bound and will not migrate in the first place. Both changes ship forward-only.

**Phases completed:**
- Phase 1 — Defensive recovery in `OidcClient.token()`: adds inner `do/catch let error as EncryptorError` — deletes corrupted ciphertext, falls through to re-auth; adds `import PingStorage`
- Phase 2 — `Keychain.save()` split into separate `deleteQuery` (primary key only) and `addQuery` with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` to prevent iCloud backup migration on new saves
- Phase 3 — Tests: `ThrowingMockStorage`, `testTokenRecoveryOnDecryptionFailure`, `testSavedItemHasDeviceOnlyAccessibility`, `testUpgradePathDoesNotThrowDuplicateItem` (mandatory upgrade-path test verifying no `errSecDuplicateItem` on first save after upgrade)
- Phase 4 — `revoke()` hardened: `try?` on `config.storage.get()` replaced with explicit `do/catch` so the documented workaround actually clears the corrupted keychain entry

**Note:** `endSession(signOff:)` has a similar swallow pattern (known deferred — no impact on the authentication lockout scenario). The ordering test `testTokenRecoveryDeletesStorageBeforeReAuth` was not implemented (production code order is clearly correct; follow-up if needed).

# Checklist:

### DIFF
--- a/Oidc/Oidc/OidcClient.swift
+++ b/Oidc/Oidc/OidcClient.swift
@@ -11,8 +11,9 @@
 
 import Foundation
 import PingLogger
-import PingOrchestrate
 import PingNetwork
+import PingOrchestrate
+import PingStorage
 
 /// Class representing an OpenID Connect client.
 /// - Property pkce: PKCE  object used for the Authorization call.
@@ -115,23 +116,29 @@ public class OidcClient {
         
         config.logger.i("Getting access token")
         do {
-            if let cached = try await config.storage.get() {
-                if !cached.isExpired(threshold: config.refreshThreshold) {
-                    config.logger.i("Token is not expired. Returning cached token.")
-                    return .success(cached)
-                }
-                config.logger.i("Token is expired. Attempting to refresh.")
-                if let cachedefreshToken = cached.refreshToken {
-                    do {
-                        let refreshedToken = try await refreshToken(cachedefreshToken)
-                        return .success(refreshedToken)
-                    } catch {
-                        config.logger.e("Failed to refresh token. Revoking token and re-authenticating.", error: error)
-                        await revoke(cached)
+            do {
+                if let cached = try await config.storage.get() {
+                    if !cached.isExpired(threshold: config.refreshThreshold) {
+                        config.logger.i("Token is not expired. Returning cached token.")
+                        return .success(cached)
+                    }
+                    config.logger.i("Token is expired. Attempting to refresh.")
+                    if let cachedefreshToken = cached.refreshToken {
+                        do {
+                            let refreshedToken = try await refreshToken(cachedefreshToken)
+                            return .success(refreshedToken)
+                        } catch {
+                            config.logger.e("Failed to refresh token. Revoking token and re-authenticating.", error: error)
+                            await revoke(cached)
+                        }
                     }
                 }
+            } catch let error as EncryptorError {
+                config.logger.w("Cached token is unreadable (device migration detected). Clearing corrupted token and re-authenticating.", error: error)
+                try? await config.storage.delete()
+                // fall through to re-authenticate below
             }
-            
+
             // Authenticate the user
             guard let agent = config.agent else {
                 return .failure(OidcError.authorizeError(message: "Agent not configured"))
@@ -191,7 +198,13 @@ public class OidcClient {
     private func revoke(_ token: Token? = nil) async {
         var accessToken = token
         if accessToken == nil {
-            accessToken = try? await config.storage.get()
+            do {
+                accessToken = try await config.storage.get()
+            } catch {
+                config.logger.w("Failed to read token for revocation. Clearing storage.", error: error)
+                try? await config.storage.delete()
+                return
+            }
         }
         if let token = accessToken {
             do {
--- a/Oidc/Oidc/OidcClient.swift
+++ b/Oidc/Oidc/OidcClient.swift
@@ -133,9 +133,24 @@ public class OidcClient {
                         }
                     }
                 }
-            } catch let error as EncryptorError {
-                config.logger.w("Cached token is unreadable (device migration detected). Clearing corrupted token and re-authenticating.", error: error)
-                try? await config.storage.delete()
+            } catch let error where error is EncryptorError || error is DecodingError {
+                // The cached bytes are unreadable — either the Secure Enclave key did not
+                // migrate with an iCloud/Quick Start device transfer (EncryptorError), or the
+                // stored payload is corrupt / from an incompatible version (DecodingError).
+                // Clear the unreadable token and fall through to re-authenticate. Other errors
+                // (e.g. errSecInteractionNotAllowed while the device is locked) are transient
+                // and must NOT delete a potentially valid token — they propagate to the outer
+                // catch unchanged.
+                config.logger.w("Cached token is unreadable (device migration or corrupt payload). Clearing corrupted token and re-authenticating.", error: error)
+                do {
+                    try await config.storage.delete()
+                } catch {
+                    // If we cannot clear the corrupted token, re-authenticating would persist a
+                    // new token over (or alongside) the unreadable one and the next token() call
+                    // would loop on the same failure. Surface a real error instead of looping.
+                    config.logger.e("Failed to clear unreadable token. Aborting to avoid a silent retry loop.", error: error)
+                    return .failure(OidcError.authorizeError(cause: error))
+                }
                 // fall through to re-authenticate below
             }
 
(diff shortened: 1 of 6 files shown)


Storage/Storage/KeychainStorage.swift
```diff
@@ -30,18 +30,32 @@ public actor Keychain<T: Codable & Sendable>: Storage {
...
         ] as [String: Any]
-        
-        query[kSecValueData  as String] = try await encryptor.encrypt(data: data)
-        
-        SecItemDelete(query as CFDictionary) // Remove any existing item
-        let status = SecItemAdd(query as CFDictionary, nil)
-        
+        addQuery[kSecValueData as String] = try await encryptor.encrypt(data: data)
+
+        let status = SecItemAdd(addQuery as CFDictionary, nil)
```
vahancouver, line 64: The refactor reversed the order of operations. The original code encrypted first (so a failed encrypt left the old token intact); the new code deletes first:
  
```
  SecItemDelete(deleteQuery as CFDictionary)           // old token gone
  addQuery[kSecValueData] = try await encryptor.encrypt(data: data)  // if this throws...
  SecItemAdd(addQuery as CFDictionary, nil)             // never reached —  keychain is empty
```

  If encrypt() throws after the delete (e.g., SE key temporarily unavailable under lock/pressure), the keychain slot is empty. Fix: move SecItemDelete to after encrypt() succeeds.

Oidc/Oidc/OidcClient.swift
```diff
@@ -115,23 +116,29 @@ public class OidcClient {
...
+                            config.logger.e("Failed to refresh token. Revoking token and re-authenticating.", error: error)
+                            await revoke(cached)
+                        }
                     }
                 }
+            } catch let error as EncryptorError {
+                config.logger.w("Cached token is unreadable (device migration detected). Clearing corrupted token and re-authenticating.", error: error)
+                try? await config.storage.delete()
+                // fall through to re-authenticate below
             }
```
vahancouver, line 155: infinite silent failure loop when delete()  itself fails

  ```
} catch let error as EncryptorError {
      try? await config.storage.delete()   // failure silently discarded
      // fall through to re-authenticate
  }
```

  If delete() throws (e.g., errSecInteractionNotAllowed while device is locked), the corrupted item stays. The next token() call hits EncryptorError again and the cycle repeats forever — the caller never gets a useful error. Fix: propagate a failure if delete cannot clear the corrupted token.

File: Oidc/Oidc/OidcClient.swift [hunk 3]
```diff
@@ -191,7 +198,13 @@ public class OidcClient {
     private func revoke(_ token: Token? = nil) async {
         var accessToken = token
         if accessToken == nil {
-            accessToken = try? await config.storage.get()
+            do {
+                accessToken = try await config.storage.get()
+            } catch {
```
vahancouver: catch-all destroys valid tokens on transient  errors

 ```
 } catch {   // any error, including errSecInteractionNotAllowed
      try? await config.storage.delete()
      return
  }
```

  This fires on transient errors (device locked mid-Background-App-Refresh), permanently deleting a perfectly valid token without revoking the server session. Should scope deletion to EncryptorError only; other errors should skip the revoke without deleting.

File: Oidc/Oidc/OidcClient.swift [hunk 4]
```diff
@@ -115,23 +116,29 @@ public class OidcClient {
...
+                        do {
+                            let refreshedToken = try await refreshToken(cachedefreshToken)
+                            return .success(refreshedToken)
+                        } catch {
+                            config.logger.e("Failed to refresh token. Revoking token and re-authenticating.", error: error)
+                            await revoke(cached)
+                        }
                     }
                 }
+            } catch let error as EncryptorError {
```
vahancouver: non-EncryptorError from storage.get() is  misclassified

  Other errors from storage.get() (e.g., KeychainError.unableToRetrieve, JSON decoding errors) escape the inner catch and surface as a misleading authorizeError, with no recovery attempted and the corrupted token left in place.

vahancouver (commented):
The fix addresses the reported issue. However there might be other edge cases.
Please see my inline comments. 

Plus we we could add tests for those cases
  - No test for delete() throwing during recovery (ThrowingMock has no
  throwOnDelete flag, so the infinite loop is untestable)
  - No test for non-EncryptorError from storage.get()
  - No test for revoke() with a transient storage error
  - No test verifying the old token survives a failed save() (which would catch
  the data loss regression)

Storage/Storage/KeychainStorage.swift
george-bafaloukas-forgerock: Good catch — fixed in 73798ee. `save()` now encrypts first and only deletes the existing item once encryption succeeds, so a transient encrypt failure (SE key unavailable under lock/pressure) can never leave the keychain slot empty. Added `testExistingTokenSurvivesFailedSave` to lock this in.

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: Fixed in 73798ee. The recovery path no longer swallows the delete failure with `try?` — if `delete()` throws, `token()` now returns `.failure(OidcError.authorizeError(cause:))` instead of falling through to re-auth, so the caller gets a real error rather than looping on the same corrupted token. Covered by `testTokenRecoveryFailsWhenDeleteFails`.

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: Fixed in 73798ee. `revoke()` now scopes deletion to `EncryptorError`/`DecodingError` only. A transient error (e.g. `errSecInteractionNotAllowed` while locked) logs a warning and returns without deleting, leaving the potentially-valid token intact. Covered by `testRevokeLeavesTokenIntactOnTransientStorageError`.

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: Fixed in 73798ee. The inner catch now recovers from both `EncryptorError` and `DecodingError` (corrupt/incompatible stored payload) — both mean the bytes are unreadable, so we clear + re-auth. All other errors (transient keychain failures) still propagate to the outer catch without deleting. Covered by `testTokenDoesNotRecoverOnTransientStorageError`.

george-bafaloukas-forgerock:
Thanks for the thorough review @vahancouver — all four points were valid (two were regressions this PR introduced). Addressed in 73798ee:

| Comment | Fix |
|---|---|
| `save()` delete-before-encrypt data loss | Encrypt first, then delete, then add — slot is never left empty on a failed encrypt |
| Infinite loop when `delete()` fails | Return `.failure` instead of falling through to re-auth |
| `revoke()` catch-all deletes valid tokens | Scope deletion to `EncryptorError`/`DecodingError`; transient errors leave the token intact |
| non-`EncryptorError` from `get()` misclassified | Recover from `DecodingError` too; transient errors still propagate without deletion |

Added all four tests you suggested:
- `testExistingTokenSurvivesFailedSave` (data-loss regression guard)
- `testTokenRecoveryFailsWhenDeleteFails` (infinite-loop guard)
- `testTokenDoesNotRecoverOnTransientStorageError`
- `testRevokeLeavesTokenIntactOnTransientStorageError`

`ThrowingMock` gained a `throwOnDelete` flag and a configurable `getError` to drive these. Full run: 57/57 OidcTests + 6/6 KeychainStorageTests green. Ready for another look.

vahancouver (approved):
LGTM

Oidc/Oidc/OidcClient.swift
```diff
@@ -11,8 +11,9 @@
 
 import Foundation
 import PingLogger
-import PingOrchestrate
 import PingNetwork
+import PingOrchestrate
+import PingStorage
```
spetrov, line 16: Not entirely sure if this is needed, but maybe we should add `PingStorage` to the `PingOidc` as direct dependency in `Package.swift` and add `s.ios.dependency 'PingStorage', '~> 2.0.0'` to `PingOidc.podspec`...?

spetrov (approved):
LGTM!

Oidc/Oidc/OidcClient.swift
george-bafaloukas-forgerock: The import is needed.

[OidcClient.swift:136](vscode-webview://1f1r1qifcq5rs3nttur9kjbkldi0ug3peom65dcmu61s98fg6jl4/Oidc/Oidc/OidcClient.swift#L136) and [:218](vscode-webview://1f1r1qifcq5rs3nttur9kjbkldi0ug3peom65dcmu61s98fg6jl4/Oidc/Oidc/OidcClient.swift#L218) both reference EncryptorError, which is defined in the PingStorage module ([Storage/Storage/SecuredKeyEncryptor.swift:55](vscode-webview://1f1r1qifcq5rs3nttur9kjbkldi0ug3peom65dcmu61s98fg6jl4/Storage/Storage/SecuredKeyEncryptor.swift#L55)). Without import PingStorage, those error is EncryptorError checks won't compile.
We could add it to the pods and SPM but it is not strictly required:
**PingOidc already gets PingStorage transitively:
PingOidc → PingOrchestrate → PingStorage**
Both SPM and CocoaPods resolve transitive modules, so import PingStorage and EncryptorError compile today without a direct declaration. 

I would say we address dependencies and other such fixes in a separate ticket for Tech Debt

PR #168, opened 2026-06-22
Sample project stale reference of ignored file

george-bafaloukas-forgerock:
Project does not build unless you have a configs folder/file that should be optional and is in the gitignore


spetrov:
This issue has been resolved with #169.

PR #170, opened 2026-06-23
Sdks 5143 image component

vahancouver:
[SDKS-5143](https://pingidentity.atlassian.net/browse/SDKS-5143) [iOS] Add image-aware Forms response support
Simulator Screenshot - iPhone Air - 2026-06-23 at 10 57 36

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying images in DaVinci forms.
  * Images can include descriptions and optional secure hyperlinks.
  * Added loading, unavailable, and failure states for image content.
* **Bug Fixes**
  * Improved accessibility, sizing, and presentation of displayed images.
* **Tests**
  * Added coverage for image parsing, optional links, defaults, and display behavior.
* **Chores**
  * Excluded generated test result files from version control.


Davinci/Davinci/collector/ImageCollector.swift
```diff
@@ -0,0 +1,59 @@
...
+import PingDavinciPlugin
+
+/// A display-only collector that surfaces an image in DaVinci authentication flows.
+///
+/// The server sends the image as a URL in the `imageUrl` field. An optional
+/// `hyperlinkUrl` field may be present to provide a tappable link associated
+/// with the image.
+///
+/// This collector does not participate in form submission — `payload()` always returns `nil`.
+public class ImageCollector: Collector, @unchecked Sendable {
```
vibhorgoswami, line 21: Does iOS have a Collector option similar to Android? `Collector<Nothing>`?

vibhorgoswami (approved):
LGTM! Please update the screenshot to show the border around the image.

witrisna (approved):

spetrov (approved):
LGTM! but let's merge this after the 2.1 release, please.

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
```diff
@@ -0,0 +1,85 @@
...
+/// hyperlink. This view converts the URL string to a `URL` locally and displays
+/// the image using `AsyncImage`. If the URL is absent or malformed a placeholder
+/// icon is shown instead. When `hyperlinkUrl` is non-nil and non-empty, a tappable
+/// `Link` is shown below the image.
+struct ImageView: View {
+    let collector: ImageCollector
+
+    var body: some View {
+        VStack(spacing: 12) {
+            if let imageURL = URL(string: collector.imageUrl) {
```
george-bafaloukas-forgerock: [Low] `imageUrl` guard is now inconsistent with the rest of this view. This commit added `!collector.description.isEmpty` (L62) and `!hyperlinkUrl.isEmpty` (L70) guards, but the image branch still keys solely off `URL(string: collector.imageUrl)`.

That matters because the behavior of an empty string is toolchain-split: on the modern swift-foundation parser (iOS 17+, what CI/sims use) `URL(string: "")` is `nil` → correctly falls to the `else` placeholder; but this app's min deployment is **iOS 16.0**, whose NSURL parser returns a *non-nil* empty URL → the `AsyncImage` branch is taken and can sit on `.empty` (`ProgressView` spinning forever).

Mirror the hyperlink guard for deterministic behavior across the supported range:
```swift
if !collector.imageUrl.isEmpty, let imageURL = URL(string: collector.imageUrl) {
```

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
```diff
@@ -0,0 +1,85 @@
...
+                Text(collector.description)
+                    .font(.subheadline)
+                    .multilineTextAlignment(.center)
+                    .padding(.horizontal)
+            }
+
+            if let hyperlinkUrl = collector.hyperlinkUrl,
+               !hyperlinkUrl.isEmpty,
+               let linkURL = URL(string: hyperlinkUrl) {
+                Link(hyperlinkUrl, destination: linkURL)
```
george-bafaloukas-forgerock: [Medium] this opens a server-controlled `hyperlinkUrl` with no scheme allowlist. `hyperlinkUrl` is a raw server-provided string handed straight to `URL(string:)` and a SwiftUI `Link`; on tap, `openURL` routes non-`http(s)` schemes to the system/app URL handler. This app's `Info.plist` registers its own deeplink/redirect schemes (`oauth2redirect`, `frauth`, `myapp`, the Google/Facebook client-ID schemes), so a malicious or misconfigured DaVinci flow could emit e.g. `oauth2redirect://…` and re-enter the app's OAuth-redirect handling out of band of the flow.

The risk is gated behind a trusted-server boundary and this is sample code — hence Medium, not High — but the sample is the reference integration people copy, so an allowlist is worth teaching here:
```swift
if let hyperlinkUrl = collector.hyperlinkUrl,
   !hyperlinkUrl.isEmpty,
   let linkURL = URL(string: hyperlinkUrl),
   let scheme = linkURL.scheme?.lowercased(),
   scheme == "http" || scheme == "https" {
    Link(hyperlinkUrl, destination: linkURL)
}
```
The SDK `ImageCollector` itself is fine — it only stores the string; this is purely a sample-app rendering concern.

.gitignore
```diff
@@ -93,6 +93,8 @@ AGENTS.md
 *.orig
 **/* copy.xcodeproj/
 /SourcePackages
+/.polaris
+/TestResults.xcresult
```
george-bafaloukas-forgerock, line 94: [nit] These two ignores (`/.polaris`, `/TestResults.xcresult`) are unrelated to SDKS-5143 and look like scope-creep in a feature PR. `/.polaris` is contributor-tooling-specific and is better placed in a personal global `core.excludesFile` than committed repo-wide; `/TestResults.xcresult` is reasonable but, if wanted, it'd be cleaner to land it in its own housekeeping change. Not blocking.

Davinci/DavinciTests/ImageCollectorTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        let originalDescription = collector.description
+        let originalHyperlinkUrl = collector.hyperlinkUrl
+
+        collector.initialize(with: "some override value")
+
+        XCTAssertEqual(collector.key, originalKey)
+        XCTAssertEqual(collector.imageUrl, originalImageUrl)
+        XCTAssertEqual(collector.description, originalDescription)
+        XCTAssertEqual(collector.hyperlinkUrl, originalHyperlinkUrl)
+    }
```
george-bafaloukas-forgerock, line 127: [Low] Coverage of `init`/defaults/`id`/`payload()`/`initialize()` is solid and on par with the sibling suites. One gap worth closing: nothing asserts that a JSON node with `type`/`inputType` `"IMAGE"` actually resolves to an `ImageCollector` via `CollectorFactory.collector(daVinci:from:)`. That path falls through silently when no closure is registered, so a registration regression in `DaVinci.swift` wouldn't be caught here.

Suggest one factory-level test (register via `CollectorFactory.shared`, build from an `["type": "IMAGE", ...]` array, assert the element is an `ImageCollector`), with `await CollectorFactory.shared.reset()` in `setUp` to avoid cross-test leakage. Same gap exists in most sibling suites, so this is opportunistic rather than blocking.

george-bafaloukas-forgerock (commented):
Overall really good one Vahan. I have some small comments on, maybe worth discussing fixing as we are in no rush to merge that in

Davinci/Davinci/collector/ImageCollector.swift
george-bafaloukas-forgerock: Yes — the iOS analog of Android's `Collector<Nothing>` is the associated type `T = Never`, surfaced as `payload() -> Never?`. The `Collector` protocol is generic over `T` (the payload type) and `Never` is Swift's bottom type, so `Never?` is the idiomatic "this collector never produces a payload" signature. It's the established pattern for all display-only collectors in this module — `LabelCollector`, `ReadOnlyTextCollector`, and `QRCodeCollector` all declare `payload() -> Never?` likewise. So `ImageCollector` here is already doing the equivalent of `Collector<Nothing>`; no extra option is needed.

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
vahancouver: added

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
vahancouver: added

.gitignore
vahancouver: removed /.polaris

Davinci/DavinciTests/ImageCollectorTests.swift
vahancouver: added the missing test

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
```diff
@@ -0,0 +1,87 @@
...
+                    .multilineTextAlignment(.center)
+                    .padding(.horizontal)
+            }
+
+            if let hyperlinkUrl = collector.hyperlinkUrl,
+               !hyperlinkUrl.isEmpty,
+               let linkURL = URL(string: hyperlinkUrl),
+               let scheme = linkURL.scheme?.lowercased(),
+               scheme == "http" || scheme == "https" {
+                Link(hyperlinkUrl, destination: linkURL)
```
george-bafaloukas-forgerock: [Optional] The first argument to `Link` is the label VoiceOver reads aloud — passing the raw URL string (`hyperlinkUrl`) gives users `"https://acme.com/tc?locale=en&ver=3"` as the tap target text. `collector.description` is already parsed and is the natural accessible label:
```swift
Link(collector.description.isEmpty ? "Open link" : collector.description, destination: linkURL)
```

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
```diff
@@ -0,0 +1,87 @@
...
+                }
+            } else {
+                Image(systemName: "photo")
+                    .resizable()
+                    .scaledToFit()
+                    .frame(width: 120, height: 120)
+                    .foregroundColor(.secondary)
+                Text("Image unavailable")
+                    .font(.caption)
+                    .foregroundColor(.secondary)
```
george-bafaloukas-forgerock: [Optional] This `Image + Text` placeholder block is byte-for-byte identical to the `case .failure:` branch above (lines 36–43). A private `@ViewBuilder` would mean one place to update:
```swift
@ViewBuilder
private var imagePlaceholder: some View {
    Image(systemName: "photo")
        .resizable().scaledToFit()
        .frame(width: 120, height: 120)
        .foregroundColor(.secondary)
    Text("Image unavailable")
        .font(.caption).foregroundColor(.secondary)
}
```

SampleApps/PingExample/PingExample/ContinueNodeView.swift
```diff
@@ -90,6 +90,8 @@ struct ContinueNodeView: View {
                     BooleanCollectorView(field: booleanCollector, onNodeUpdated: onNodeUpdated)
                 case let readOnlyTextCollector as ReadOnlyTextCollector:
```
george-bafaloukas-forgerock, line 91: [Optional] For consistency with `QRCodeView`, `PollingView`, and `ProtectView` (which all carry `.id()` modifiers), consider:
```swift
ImageView(collector: imageCollector).id(imageCollector.id)
```

Davinci/Davinci/collector/ImageCollector.swift
```diff
@@ -0,0 +1,59 @@
...
+    // MARK: - Properties
+
+    /// The key identifying this collector within the form.
+    public private(set) var key: String = ""
+
+    /// The URL string of the image to display. Empty string if not provided by the server.
+    public private(set) var imageUrl: String = ""
+
+    /// A description or alt text for the image. Empty string if not provided by the server.
+    public private(set) var description: String = ""
```
george-bafaloukas-forgerock, line 36: [Optional] `description` shadows `CustomStringConvertible.description`. Sibling collectors use `content` (`LabelCollector`) or `fallbackText` (`QRCodeCollector`) for display-text fields. Renaming to `content` or `altText` avoids a silent collision if `CustomStringConvertible` conformance is ever added to the `Collector` hierarchy.

george-bafaloukas-forgerock (approved):
LGTM — all four prior comments have been addressed. Leaving a few optional suggestions below.

**[Optional] CHANGELOG.md** — The `[Unreleased]` section has no entry for `ImageCollector` / SDKS-5143. Per the repo's `ping-sdk:changelog-entry` convention an `#### Added` line should be included before merge, e.g.:
```
- Added `ImageCollector` to support image display in DaVinci forms [SDKS-5143]
```

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
vahancouver: Fixed

SampleApps/PingExample/PingExample/Collectors/ImageView.swift
vahancouver: Fixed

SampleApps/PingExample/PingExample/ContinueNodeView.swift
vahancouver: Fixed

Davinci/Davinci/collector/ImageCollector.swift
vahancouver: This will require changes in Android and web also if we decide to rename the property. I don't think `CustomStringConvertible` conformance will ever be added here. 

No changes made.

PR #171, opened 2026-06-23
SDKS-5068: QA for the Unified SDK Configuration

spetrov:
# JIRA Ticket

  [SDKS-5068](https://pingidentity.atlassian.net/browse/SDKS-5068) [QA] Standardize SDK Configuration

  # Description

  Added E2E test coverage for the unified JSON configuration factory functions
  `Journey.createJourney(json:)`, `DaVinci.createDaVinci(json:)`,
  `OidcWebClient.createOidcWebClient(json:)`, and `OidcDeviceClient.createOidcDeviceClient(json:)`.
  All tests are placed alongside existing integration tests in the respective modules.

  ## New test files

  - `Journey/JourneyTests/Integration Tests/JourneyJsonConfigE2ETest.swift` —
    `createJourney(json:)` happy path (minimal + full fixture), default values for all optional
    field errors (journey block, serverUrl, oidc fields), unknown field tolerance at every level,
    and wrong-type errors for timeout, serverUrl, realm, cookieName, scopes, refreshThreshold,
    and the oidc dict itself.

  - `Davinci/DavinciTests/integration tests/DaVinciJsonConfigE2ETest.swift` —
    `createDaVinci(json:)` happy path (minimal + full fixture), default timeout value, Journey-
    specific fields silently ignored, unknown field tolerance, missing required oidc field errors,
    wrong-type errors for timeout, clientId, scopes, and refreshThreshold, and four live login
    tests against the PingOne tenant (minimal, full, token retrieval with explicit assertions,
    unknown fields).

  - `Oidc/OidcTests/IntegrationTests/OidcWebClientJsonConfigE2ETests.swift` —
    `createOidcWebClient(json:)` happy path (minimal + full fixture), default timeout value,
    unknown field tolerance, missing required oidc field errors, wrong-type errors for timeout,
    oidc dict, scopes, and refreshThreshold, and four live tests against AIC that verify OIDC
    discovery runs and a well-formed PKCE authorize URL is constructed (browser step intercepted —
    no real ASWebAuthenticationSession opens).

  - `Davinci/DavinciTests/integration tests/DeviceClientJsonConfigE2ETest.swift` —
    `createOidcDeviceClient(json:)` default log level check, full JSON log level application,
    unknown field tolerance, missing required oidc field errors, wrong-type errors for oidc dict,
    scopes, and openId dict, and three live device authorization flow tests against the PingOne
    tenant (minimal, full JSON including openId endpoint override, unknown fields). Flow is
    cancelled after the first `.started` emission — no user interaction required.

### DIFF
--- a/Davinci/Davinci.xcodeproj/project.pbxproj
+++ b/Davinci/Davinci.xcodeproj/project.pbxproj
@@ -14,9 +14,6 @@
 		3A292C822BF58462006977EF /* Agent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3A292C812BF58462006977EF /* Agent.swift */; };
 		3A5441AA2BCDF20700385131 /* PingDavinci.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3A5441A12BCDF20700385131 /* PingDavinci.framework */; };
 		3A5441AF2BCDF20700385131 /* DaVinciTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3A5441AE2BCDF20700385131 /* DaVinciTests.swift */; };
-		SDKS50660003000000000002 /* DaVinciJsonConfigTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = SDKS50660003000000000001 /* DaVinciJsonConfigTests.swift */; };
-		AA4785031F0000000000AA01 /* DaVinciDeviceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA4785021F0000000000AA01 /* DaVinciDeviceTests.swift */; };
-		25554859C34DBAACE79ABA36 /* DaVinciPARTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2AD0544A9787573F4A02A278 /* DaVinciPARTests.swift */; };
 		3A5441B02BCDF20700385131 /* Davinci.h in Headers */ = {isa = PBXBuildFile; fileRef = 3A5441A42BCDF20700385131 /* Davinci.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		3A57CB622C44D68C001EC9A0 /* User.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3A57CB612C44D688001EC9A0 /* User.swift */; };
 		3AC13E352C3FFF1000DEF23A /* Form.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3AC13E322C3FFF1000DEF23A /* Form.swift */; };
@@ -87,6 +84,9 @@
 		ECE2B8712DB7C36800E2C51E /* PhoneNumberCollectorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECE2B8702DB7C35400E2C51E /* PhoneNumberCollectorTests.swift */; };
 		ECF053222D9C31A3004C9AAE /* DeviceRegistrationCollector.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECF053212D9C319C004C9AAE /* DeviceRegistrationCollector.swift */; };
 		ECF053242D9C4275004C9AAE /* Device.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECF053232D9C4273004C9AAE /* Device.swift */; };
+		SDKS50660003000000000002 /* DaVinciJsonConfigTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = SDKS50660003000000000001 /* DaVinciJsonConfigTests.swift */; };
+		SDKS50680002000000000002 /* DaVinciJsonConfigE2ETest.swift in Sources */ = {isa = PBXBuildFile; fileRef = SDKS50680002000000000001 /* DaVinciJsonConfigE2ETest.swift */; };
+		SDKS50680004000000000002 /* DeviceClientJsonConfigE2ETest.swift in Sources */ = {isa = PBXBuildFile; fileRef = SDKS50680004000000000001 /* DeviceClientJsonConfigE2ETest.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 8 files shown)


Davinci/Davinci.xcodeproj/project.pbxproj
```diff
@@ -14,9 +14,6 @@
 		3A292C822BF58462006977EF /* Agent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3A292C812BF58462006977EF /* Agent.swift */; };
 		3A5441AA2BCDF20700385131 /* PingDavinci.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3A5441A12BCDF20700385131 /* PingDavinci.framework */; };
 		3A5441AF2BCDF20700385131 /* DaVinciTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3A5441AE2BCDF20700385131 /* DaVinciTests.swift */; };
-		SDKS50660003000000000002 /* DaVinciJsonConfigTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = SDKS50660003000000000001 /* DaVinciJsonConfigTests.swift */; };
-		AA4785031F0000000000AA01 /* DaVinciDeviceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA4785021F0000000000AA01 /* DaVinciDeviceTests.swift */; };
-		25554859C34DBAACE79ABA36 /* DaVinciPARTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2AD0544A9787573F4A02A278 /* DaVinciPARTests.swift */; };
```
vahancouver, line 19: DaVinciDeviceTests.swift and DaVinciPARTests.swift are removed here. Did you mean to do this?

Davinci/DavinciTests/integration tests/DeviceClientJsonConfigE2ETest.swift
```diff
@@ -0,0 +1,268 @@
...
+            "log": "DEBUG",
+            "oidc": [
+                "clientId": config.deviceClientId,
+                "discoveryEndpoint": config.discoveryEndpoint,
+                "scopes": ["openid", "email", "address", "phone", "profile"],
+                "redirectUri": config.redirectUri,
+                "refreshThreshold": 60,
+                "openId": [
+                    "deviceAuthorizationEndpoint":
+                        "https://auth.pingone.ca/300c4f2a-39d4-4ba9-a18a-f6de246006f4/as/device_authorization"
```
vahancouver, line 66: should this come from the config as weel?

Oidc/OidcTests/IntegrationTests/OidcWebClientJsonConfigE2ETests.swift
```diff
@@ -0,0 +1,301 @@
...
+
+    // Triggers OIDC discovery via a JSON-built client and asserts the resulting authorize URL uses the AIC server and contains required PKCE parameters.
+    func testOpenIdEndpointsAreLoaded() async throws {
+        let web = try OidcWebClient.createOidcWebClient(json: minimalJson).get()
+        do { _ = try await web.authorize() } catch { /* token exchange fails with fake code — expected */ }
+
+        let authorizeURL = try XCTUnwrap(browser.launchedURL,
+                                         "Browser was not launched — discovery may have failed")
+        XCTAssertTrue(
+            authorizeURL.host?.contains("openam-sdks.forgeblocks.com") == true,
```
vahancouver: Can you double check if this was supposed to be harcoded?

vahancouver (changes requested):
Left minor comments

Davinci/DavinciTests/integration tests/DeviceClientJsonConfigE2ETest.swift
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    // MARK: - Config parsing: field values
+
+    // Verifies that omitting optional fields does not produce a WarningLogger (i.e. the default log level is the standard logger).
+    func testMinimalJsonConfig_defaultValues() throws {
+        let client = try OidcDeviceClient.createOidcDeviceClient(json: minimalJson).get()
+        XCTAssertFalse(client.logger is WarningLogger,
+                       "Default log level should not be WarningLogger — expected the standard logger")
+    }
```
george-bafaloukas-forgerock, line 79: Nice coverage here! One subtle gotcha worth flagging: this default-logger assertion can be order-dependent. `WorkflowConfig.logger`'s `didSet` propagates to the global `LogManager.logger`, and `JsonConfigParser.logLevel()` falls back to that global when no `"log"` key is present. In the same DavinciTests target, `DaVinciJsonConfigTests.testCreateDaVinci_logMapping_error` builds a config with `log=ERROR`, which flips the global to `WarningLogger`. If that test runs before this one, `XCTAssertFalse(client.logger is WarningLogger)` could fail even though the parser is behaving correctly.

A couple of options to make it deterministic:
- Reset `LogManager.logger = NoneLogger()` in `setUp`/`tearDown`, or
- Assert the concrete expected type directly (the real default is actually `NoneLogger`, which the current loose `is not WarningLogger` check happens to allow — so the message "expected the standard logger" is a little misleading too).

Totally minor, just want to spare you a confusing intermittent red on CI. 🙂

Davinci/DavinciTests/integration tests/DaVinciJsonConfigE2ETest.swift
```diff
@@ -0,0 +1,351 @@
...
+        XCTAssertEqual(field, "oidc.refreshThreshold")
+    }
+
+    // MARK: - Live login
+
+    // Builds a DaVinci instance from minimal JSON and drives the full two-step login form against the live PingOne tenant.
+    func testLogin_withMinimalJsonConfig_succeeds() async throws {
+        let result = DaVinci.createDaVinci(json: minimalJson)
+        guard case .success(let daVinci) = result else {
+            XCTFail("createDaVinci(json:) failed: \(result)")
```
george-bafaloukas-forgerock, line 241: Small consistency thought: the MFADeviceTests changes in this same PR deliberately swap `as!` force-casts for `guard ... else { throw XCTSkip(...) }` so the suite degrades gracefully when the tenant is unavailable. These new live tests (here and the device-flow ones) instead use `XCTFail`/`XCTUnwrap` on the same kind of environment dependency, so they'll hard-fail CI when the tenant is down or credentials rotate — which is exactly the scenario the MFADevice change was guarding against.

No strong preference on which way to go, but it'd be great to land on one policy across the PR since these all run in `PingExample.xctestplan`. If `XCTSkip`-on-environment-failure is the intended direction, applying it here too would keep things consistent. 👍

Davinci/Davinci.xcodeproj/project.pbxproj
spetrov: The develop branch has duplicate `PBXBuildFile` and `PBXFileReference` entries for those two files (each appears twice). When Xcode added the new test files for this PR it cleaned up the duplicates. Both `DaVinciDeviceTests.swift` and `DaVinciPARTests.swift` are still fully registered in all four required pbxproj sections - no files were removed.

Davinci/DavinciTests/integration tests/DeviceClientJsonConfigE2ETest.swift
spetrov: The hardcoded value here is intentional — the purpose of `testSucceedsWithFullJsonConfig` is specifically to exercise the `openId.deviceAuthorizationEndpoint` override path: we want the JSON parser to honour a custom endpoint URL rather than the one returned by discovery. Pulling it from config would give us the same URL we'd get from discovery and wouldn't actually test the override.

Oidc/OidcTests/IntegrationTests/OidcWebClientJsonConfigE2ETests.swift
spetrov: You're right, that should come from the config. Replaced the hardcoded host with a value coming from `aicConfig.discoveryEndpoint`

Davinci/DavinciTests/integration tests/DeviceClientJsonConfigE2ETest.swift
spetrov: Agreed - good catch. 
Fixed in two parts: changed the assertion to `XCTAssertTrue(client.logger is NoneLogger)` which pins the actual expected type rather than ruling out a single wrong type, and added `LogManager.logger = NoneLogger()` resets to both `setUp` and `tearDown` so any test in the same run that flips the global logger can't affect this one.

Davinci/DavinciTests/integration tests/DaVinciJsonConfigE2ETest.swift
spetrov: Agreed, applied consistently across both files - 
In `DaVinciJsonConfigE2ETest`, all XCTFail/return pairs guarding network/node results in `driveLogin` and the inline token retrieval test are replaced with `throw XCTSkip("... - environment may be unavailable")`. 
In `DeviceClientJsonConfigE2ETest`, the three live device flow tests replace `XCTUnwrap/XCTAssertTrue(gotStarted)` with the same `guard … else { throw XCTSkip(…) }` pattern. 
Config parse failures keep `XCTFail` since those are never environment-dependent.

vahancouver (approved):
LGTM

PR #172, opened 2026-06-26
Surface distinct error for Push Number Challenge

vahancouver:
[SDKS-5115](https://pingidentity.atlassian.net/browse/SDKS-5115)
[iOS] Surface distinct wrong-number error for Push Number Challenge

### DIFF
--- a/Push/Push/Core/PingAMPushResponder.swift
+++ b/Push/Push/Core/PingAMPushResponder.swift
@@ -184,6 +184,19 @@ public final class PingAMPushResponder: @unchecked Sendable {
             let response = try await httpClient.request(request: request)
             
             guard response.status.isSuccess() else {
+                if response.status == 400
+                    && approve
+                    && notification.pushType == .challenge {
+                    var message = "Number challenge failed."
+                    if let data = response.body,
+                       let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
+                       let amMessage = json["message"] as? String, !amMessage.isEmpty {
+                        message = amMessage
+                    }
+                    logger?.e(message, error: nil)
+                    throw PushError.pushNumberChallengeError(message)
+                }
+
                 let bodyDescription = response.body.flatMap { String(data: $0, encoding: .utf8) } ?? "no response body"
                 let message = "Authentication response failed with status code \(response.status): \(bodyDescription)"
                 logger?.e(message, error: nil)
--- a/Push/Push/PushError.swift
+++ b/Push/Push/PushError.swift
@@ -125,9 +125,12 @@ public enum PushError: Error, LocalizedError, Sendable {
     case registrationFailed(String)
     
     // MARK: - Network Errors
-    
+
     /// A network operation failed.
     case networkFailure(String, Error?)
+
+    /// The selected number was incorrect in a Push Number Challenge.
+    case pushNumberChallengeError(String)
     
     // MARK: - LocalizedError Conformance
     
@@ -198,6 +201,9 @@ public enum PushError: Error, LocalizedError, Sendable {
                 return "Network failure: \(message) - \(underlyingError.localizedDescription)"
             }
             return "Network failure: \(message)"
+
+        case .pushNumberChallengeError(let message):
+            return message
         }
     }
 }
--- a/Push/Push/PushError.swift
+++ b/Push/Push/PushError.swift
@@ -129,7 +129,7 @@ public enum PushError: Error, LocalizedError, Sendable {
     /// A network operation failed.
     case networkFailure(String, Error?)
 
-    /// The selected number was incorrect in a Push Number Challenge.
+    /// A Push Number Challenge response was rejected by the server (HTTP 400).
     case pushNumberChallengeError(String)
     
     // MARK: - LocalizedError Conformance
@@ -203,7 +203,7 @@ public enum PushError: Error, LocalizedError, Sendable {
             return "Network failure: \(message)"
 
         case .pushNumberChallengeError(let message):
-            return message
+            return "Push Number Challenge failed: \(message)"
         }
     }
 }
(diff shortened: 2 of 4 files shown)


rodrigoareis (approved):
Changes looks good to me

Push/Push/Core/PingAMPushResponder.swift
```diff
@@ -184,6 +184,19 @@ public final class PingAMPushResponder: @unchecked Sendable {
             let response = try await httpClient.request(request: request)
             
             guard response.status.isSuccess() else {
+                if response.status == 400
+                    && approve
+                    && notification.pushType == .challenge {
```
george-bafaloukas-forgerock, line 189: **Over-broad 400 classification.** This treats *any* HTTP 400 on an approved challenge push as a wrong-number failure, but AM can return 400 for other reasons on a challenge approval — expired/stale notification, malformed/replayed JWT, etc. Those would be mislabeled as a number-challenge failure and the user told to re-check the number when that isn't the cause.

If AM returns a distinguishable error/detail code in the body, prefer keying on that rather than assuming every `400 + approve + .challenge` is a wrong-number event. At minimum, align the case doc comment (see `PushError.swift`) so it reflects challenge-response rejections generally rather than strictly a wrong number.

Push/Push/PushError.swift
```diff
@@ -125,9 +125,12 @@ public enum PushError: Error, LocalizedError, Sendable {
     case registrationFailed(String)
     
     // MARK: - Network Errors
-    
+
     /// A network operation failed.
     case networkFailure(String, Error?)
+
+    /// The selected number was incorrect in a Push Number Challenge.
```
george-bafaloukas-forgerock: **Doc/behavior mismatch.** The comment says "The selected number was incorrect," but the throwing site fires on *any* 400 for an approved challenge (see `PingAMPushResponder.swift`), not strictly a wrong number. Consider softening to something like "A Push Number Challenge response was rejected by the server" so the doc doesn't over-promise number-specific semantics.

Push/Push/PushError.swift
```diff
@@ -125,9 +125,12 @@ public enum PushError: Error, LocalizedError, Sendable {
...
+    case pushNumberChallengeError(String)
```
george-bafaloukas-forgerock, line 133: **Source-compatibility note (informational).** `PushError` is a public, non-`@frozen` enum, so adding a case forces any client doing an exhaustive `switch` over `PushError` to add a branch on recompile. Consistent with how this enum already evolves, so acceptable — just worth calling out in the release notes / CHANGELOG so integrators aren't surprised.

Push/Push/PushError.swift
```diff
@@ -198,6 +201,9 @@ public enum PushError: Error, LocalizedError, Sendable {
                 return "Network failure: \(message) - \(underlyingError.localizedDescription)"
             }
             return "Network failure: \(message)"
+
+        case .pushNumberChallengeError(let message):
+            return message
```
george-bafaloukas-forgerock: **Consistency + trust.** Every other case prefixes context ("Network failure: …", "Policy violation: …"); this one returns the raw, server-controlled AM string verbatim. That string (e.g. "Number challenge predicate not met.") is technical and non-localized — poor to surface directly in UI. Consider owning the user-facing/localized string in the SDK and keeping the raw AM message to the logger only.

Push/PushTests/PingAMPushResponderTests.swift
```diff
@@ -528,6 +528,130 @@ final class PingAMPushResponderTests: XCTestCase {
         }
     }
 
+    func testSendAuthenticationWrongNumberThrowsNumberChallengeRejected() async {
```
george-bafaloukas-forgerock: Two test notes:

1. **Coverage gap:** there's no assertion that a **non-400** error on an approved challenge (e.g. 401/403/500) falls through to `networkFailure`. That's exactly the `== 400` boundary most likely to regress if someone later widens the check to `isClientError()`. Worth adding a case.
2. **Naming (cosmetic):** tests are named `…NumberChallengeRejected` while the error is `pushNumberChallengeError` — aligning would aid grep-ability.

george-bafaloukas-forgerock (commented):
Automated code review — a few suggestions inline. Nothing blocks merge, but two items are worth addressing before merge:

**1. Missing CHANGELOG entry (project convention).** `CHANGELOG.md` logs every change under `[Unreleased]` with its `[SDKS-####]` tag. This PR adds a **public API** case (`PushError.pushNumberChallengeError`) but has no entry. Please add one under `#### Added`, e.g.:
```
- Added `PushError.pushNumberChallengeError` to surface a distinct wrong-number failure for Push Number Challenge [SDKS-5115]
```

**2. Reconsider the blanket 400 classification** — see inline comment on `PingAMPushResponder.swift`.

Overall: focused, well-tested, low-risk change that does what the ticket asks. The rest of the inline notes are polish.

spetrov (approved):
LGTM!

Push/Push/PushError.swift
vahancouver: Softened doc comment — no longer says "selected number was incorrect", now "rejected by the server (HTTP 400)"

Push/Push/PushError.swift
vahancouver: errorDescription now prefixes with "Push Number Challenge failed: " instead of returning the raw AM  string

Push/PushTests/PingAMPushResponderTests.swift
vahancouver: - Renamed 3 test methods to …PushNumberChallengeError to match the error case name
- Added testSendAuthenticationNon400OnApprovedChallengeThrowsNetworkFailure — verifies the 400 boundary  doesn't swallow 500s

Push/Push/PushError.swift
vahancouver: Added PushError.pushNumberChallengeError entry under [Unreleased] → #### Added

george-bafaloukas-forgerock (approved):
LGTM

PR #173, opened 2026-07-06
SDKS-5212 Add preferImmediatelyAvailableCredentials option to FIDO authentication

george-bafaloukas-forgerock:
# JIRA Ticket

[SDKS-5212](https://bugster.forgerock.org/jira/browse/SDKS-5212)

# Description

Adds a `preferImmediatelyAvailableCredentials` option to FIDO authentication, restricting the ceremony to locally-available credentials (no QR / nearby-device fallback), matching the legacy `FRWebAuthnManager.signInWith(preferImmediatelyAvailableCredentials:)` behavior.

Exposed on:
- `Fido.authenticate` (core)
- `FidoAuthenticationCollector.authenticate` (DaVinci)
- `FidoAuthenticationCallback.authenticate` (Journey)

Defaults to `false` to preserve existing behavior. When `true`, the security-key (cross-platform) request is not built, since a hardware key can never be "immediately available" on the local device, and `performRequests(options: .preferImmediatelyAvailableCredentials)` is used to suppress the QR/nearby-device fallback UI.

Also includes an unrelated pre-existing bug fix (`request.displayName = options.user.displayName` in `createPlatformRequest`) so the WebAuthn Registration Node's configured display name is shown during passkey creation instead of falling back to the username.

Sample app (`FidoAuthenticationCallbackView`/`FidoAuthenticationCollectorView`) exposes the new option via a "Local credentials only" toggle.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 ## [Unreleased]
 #### Fixed
 - Fixed permanent authentication failure after iCloud device migration caused by Secure Enclave key mismatch [SDKS-5172]
+- Fixed `PingFido` platform WebAuthn registration ignoring `displayName`; the display name configured on the WebAuthn Registration Node is now shown to the user during passkey creation instead of falling back to the username [TRIAGE-35655]
 
 #### Added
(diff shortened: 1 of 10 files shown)


vahancouver (approved):
LGTM

witrisna (approved):

spetrov (approved):
LGTM! 👍🏻

PR #174, opened 2026-07-13
SDKS-5156 AM/AIC backchannel auth support in PingJourney

george-bafaloukas-forgerock:
# JIRA Ticket

[SDKS-5156](https://pingidentity.atlassian.net/browse/SDKS-5156) "[iOS] AM/AIC Backchannel Authentication support in PingJourney"

# Description

Adds a native AM/AIC transactional backchannel-authentication entry point to `PingJourney`. A federation gateway initialises a transaction with AM and forwards a browser `redirectUri` to the app; the SDK extracts `authIndexType` / `authIndexValue` from that URI's query string and drives the *standard* authenticate journey to a continue / success / error node. The SDK must never call the `/am/UI/Login` browser URL in the `redirectUri`; it reconstructs the authenticate API endpoint from `JourneyConfig.serverUrl` + `JourneyConfig.realm` (realm always from config, never the URI). The change is deliberately small and reuses existing machinery: `populateRequest` already accepts an arbitrary `authIndexType`, and `NodeTransformModule` already maps 2xx/4xx/5xx responses to the correct node types — neither needed modification.

**Phases completed:**
- Phase 1 — Add `transaction` constant to `JourneyConstants`: Added `public static let transaction = "transaction"` to `JourneyConstants` (JourneyPlugin/JourneyPlugin/Constants.swift), grouped with the other auth-index constants.
- Phase 2 — Add `start(backchannelUri:configure:)` to `Journey`: Added `func start(backchannelUri: URL, configure: @Sendable (inout Options) -> Void = { _ in }) async -> Node` to `Journey/Journey/Journey.swift`, following the design doc §5.3 reference implementation exactly.
- Phase 3 — Add `JourneyBackchannelTests` (9 cases): Created `Journey/JourneyTests/JourneyBackchannelTests.swift` covering happy-path continue/success, invalid URI (missing authIndexType/authIndexValue/no query items), JourneyConfig absent, noSession flag, forceAuth flag, and AM 4xx → ErrorNode.
- Phase 4 — CHANGELOG entry + full-module build validation: Added `[Unreleased]` CHANGELOG entry; confirmed podspecs and Package.swift require no manifest edits (source globbing already covers new files).

**Design reference:** https://pingidentity.atlassian.net/wiki/spaces/SDKS/pages/2941517830

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added AM/AIC backchannel authentication support using gateway-provided redirect URIs.
  * Added options to control forced authentication and session handling.
  * Added a Backchannel authentication flow to the PingExample app, including redirect URI input and error handling.
  * Added the `transaction` Journey parameter constant.

* **Documentation**
  * Documented the backchannel authentication flow, validation behavior, and usage examples.
  * Updated the changelog for version 2.1.0.

* **Tests**
  * Added coverage for successful flows, validation errors, request options, host matching, and AM error responses.


Journey/Journey/Journey.swift
```diff
@@ -187,6 +187,45 @@ public extension Journey {
...
+        self.sharedContext.set(key: JourneyConstants.forceAuth, value: options.forceAuth)
+        self.sharedContext.set(key: JourneyConstants.noSession, value: options.noSession)
+
+        guard let components = URLComponents(url: backchannelUri, resolvingAgainstBaseURL: false),
+              let authIndexType = components.queryItems?.first(where: { $0.name == JourneyConstants.authIndexType })?.value,
+              !authIndexType.isEmpty,
+              let authIndexValue = components.queryItems?.first(where: { $0.name == JourneyConstants.authIndexValue })?.value,
+              !authIndexValue.isEmpty
+        else {
+            return FailureNode(cause: ApiError.error(400, [:], "Invalid URI or missing authIndexType/authIndexValue"))
```
witrisna, line 228: ApiError for representing an error in API call, in this case, just verify the input parameter, for Android, it may use InvalidArgumentException, something similar for iOS? and do we want to verify the url has the same host as the `serverUrl` for the Journey?

witrisna (changes requested):

Journey/Journey/Journey.swift
george-bafaloukas-forgerock: Thanks for the review — addressed the host-validation point in the latest commit: `start(backchannelUri:)` now rejects the URI if its host doesn't match `JourneyConfig.serverUrl`'s host before reading any query parameters, with a new test (`testInvalidUri_hostMismatch`) covering it.

On the `ApiError` question: kept it as-is rather than introducing a new error type. `start(_:)` and `resume(_:)` above already use `ApiError.error(400, ...)` for this same class of local pre-flight failure (missing config, invalid URI), so this keeps the new method consistent with the established convention in this file rather than introducing a second error type for the same conceptual case. Happy to revisit if you'd still like a dedicated argument-error type across the module — but wanted to flag it's a pre-existing pattern here, not something new to this PR.

witrisna: Got, if already using this pattern, keep it as is for now.

witrisna (approved):

Journey/README.md
```diff
@@ -172,6 +172,49 @@ let node = await journey.start("Login") { options in
...
+// https://<tenant>/am/UI/Login?realm=%2Falpha&authIndexType=transaction&authIndexValue=<uuid>
+let redirectUri = URL(string: gatewaySuppliedRedirectUri)!
+
+let node = await journey.start(backchannelUri: redirectUri) { options in
+    options.noSession = false
+    options.forceAuth = false
+}
+```
+
+`start(backchannelUri:configure:)` only reads the `authIndexType` and `authIndexValue` query parameters off `redirectUri` — the URI's host, path, and `realm` query parameter (if present) are ignored. The authenticate request is always sent to `JourneyConfig.serverUrl` + `JourneyConfig.realm`; the SDK never calls the `redirectUri` itself, which typically points at a browser login page rather than an API endpoint. Both `authIndexType` and `authIndexValue` must be present in the URI's query string **and** non-empty, or `start(backchannelUri:)` returns a `FailureNode(400, ...)` without making any network call — the same is true if `JourneyConfig` is missing.
```
vahancouver: this is no longer correct, the code now validates the host.

Journey/Journey/Journey.swift
```diff
@@ -187,6 +187,54 @@ public extension Journey {
...
+    ///   parameters, or `JourneyConfig` is absent.
+    func start(backchannelUri: URL, configure: @Sendable (inout Options) -> Void = { _ in }) async -> Node {
+        var options = Options()
+        configure(&options)
+        guard let journeyConfig = self.config as? JourneyConfig else {
+            return FailureNode(cause: ApiError.error(400, [:], "JourneyConfig missing"))
+        }
+
+        guard let serverHost = journeyConfig.serverUrl.flatMap({ URL(string: $0)?.host }),
+              backchannelUri.host == serverHost
```
vahancouver: hostnames are case-insensitive, .lowercased() should be used on on both sides.

Journey/Journey/Journey.swift
```diff
@@ -187,6 +187,54 @@ public extension Journey {
...
+        else {
+            return FailureNode(cause: ApiError.error(400, [:], "backchannelUri host does not match JourneyConfig.serverUrl"))
+        }
+
+        self.sharedContext.set(key: JourneyConstants.forceAuth, value: options.forceAuth)
+        self.sharedContext.set(key: JourneyConstants.noSession, value: options.noSession)
```
vahancouver: should those two be set after the guard statement in case the guard fails?

vahancouver (commented):
Left some comments and questions

Journey/README.md
george-bafaloukas-forgerock: Fixed in db9abf5 — the README now states the host is validated (must match `JourneyConfig.serverUrl`'s host, case-insensitively) instead of being ignored.

Journey/Journey/Journey.swift
george-bafaloukas-forgerock: Good catch — fixed in db9abf5. Both sides are now compared with `.lowercased()` (per RFC 3986 §3.2.2 hostnames are case-insensitive), and added `testHostMatchIsCaseInsensitive` covering an upper-cased host.

Journey/Journey/Journey.swift
george-bafaloukas-forgerock: Agreed — fixed in db9abf5. The two `sharedContext.set` calls now happen after the query-parameter guard, so a validation failure returns the `FailureNode` without leaving stale `forceAuth`/`noSession` state behind.

vahancouver (approved):
LGTM

SampleApps/PingExample/PingExample/JourneyViewModel.swift
```diff
@@ -144,6 +146,48 @@ class JourneyViewModel: ObservableObject {
...
+        guard let journey = journey else {
+            await MainActor.run {
+                self.errorMessage = "No Journey configuration found."
+                self.isLoading = false
+            }
+            return
+        }
+
+        let next = await journey.start(backchannelUri: backchannelUri) { options in
+            options.forceAuth = false
```
spetrov: These `forceAuth: false` and `noSession: false` are probably not necessary? These are already the defaults...

Package.resolved
spetrov, line 1: Are these changes in `Package.resolved` intentional? They look more like an accidental commit from your local Xcode resolve? 

AppAuth-iOS 2.0.0 --> 2.1.0, 
GoogleUtilities 8.1.0 --> 8.1.2, 
Promises 2.4.0 --> 2.4.1.

spetrov (approved):
LGTM!

PR #176, opened 2026-07-16
Add MetadataCollector for SDK Connector pause/resume model

vahancouver:
[SDKS-5142](https://pingidentity.atlassian.net/browse/SDKS-5142) 
[iOS] Add DaVinci metadata support via SDK Connector pause/resume model

## Summary by CodeRabbit

- **New Features**
  - Added METADATA step support to the DaVinci SDK Integrator flow, including resume-ready payload handling.
  - Introduced a new MetadataCollector with result/error submission, validation, and lifecycle reset behavior.
  - Updated the sample app to visualize metadata and simulate success or error outcomes.
- **Bug Fixes**
  - Improved collector-to-resume JSON generation to include METADATA action data only when a result is set.
- **Tests**
  - Added unit coverage for metadata parsing, payload lifecycle, validation/error behavior, and resume envelope generation.


Davinci/Davinci/collector/MetadataCollector.swift
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// The `eventType` value DaVinci expects on resume. Always `"action"`.
+    public func eventType() -> String {
+        return Constants.action
+    }
+
+    /// Returns a `.required` validation error when no result has been set,
+    /// or an empty array once `setResult(_:)` or `setError(...)` has been called.
+    public func validate() -> [ValidationError] {
```
george-bafaloukas-forgerock, line 133: This is the one I'd really like to flag before merge: `validate()` returns `.required` when no result has been set, and the doc comments frame this as the thing that stops the flow from continuing without an on-device result — but nothing actually calls `validate()`. The sample app's `shouldValidate(node:)` only checks for `ValidatedCollector`, and `MetadataCollector` conforms to `Validator` directly, so it's never invoked. Net effect: if an app forgets to call `setResult`/`setError` and taps the fallback "Next" button, the flow just continues with `sdkMetadata` silently omitted, rather than being blocked. Given the story here is "fail predictably and be diagnosable," I think this needs to actually be wired in (or the docs/conformance adjusted to match reality).

Davinci/Davinci/collector/MetadataCollector.swift
```diff
@@ -0,0 +1,143 @@
...
+/// }
+/// ```
+///
+/// ### Usage
+/// 1. Read `metadata` to determine which SDK operation to invoke.
+/// 2. Call `setResult(_:)` with the SDK's result, or `setError(...)` to
+///    signal the connector's client-error branch.
+/// 3. Continue the flow as usual — the SDK serialises the result and POSTs
+///    the resume envelope to DaVinci.
+public class MetadataCollector: AnyFieldCollector, Submittable, Validator, Closeable, @unchecked Sendable {
```
george-bafaloukas-forgerock, line 38: Related to the note below on `validate()` — since it's currently unreachable, the `Validator` conformance here is a bit of misleading API surface (an integrator reading this would reasonably assume the SDK enforces it client-side). Once validation is wired in this is fine as-is; otherwise might be worth a doc tweak so it's clear this isn't currently enforced.

Davinci/Davinci/collector/MetadataCollector.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// which the collector is considered ready to submit.
+    private var result: [String: Any]?
+
+    /// Creates a collector from a raw JSON component received from DaVinci.
+    ///
+    /// - Parameter json: The parsed JSON object for this `METADATA` component.
+    public required init(with json: [String: Any]) {
+        key = json[Constants.key] as? String ?? ""
+        type = json[Constants.type] as? String ?? ""
+        metadata = json[Constants.payload] as? [String: Any] ?? [:]
```
george-bafaloukas-forgerock, line 67: Small one: if `payload` comes back malformed (e.g. an array instead of an object), this silently coerces to `[:]` — indistinguishable from a legitimately-empty payload. `FidoRegistrationCollector` logs via `davinci?.config.logger` in a similar situation; might be worth a `.d`/`.w` log line here too so integrators have a breadcrumb to debug against instead of just seeing an empty dict.

DavinciPlugin/DavinciPlugin/Constants.swift
```diff
@@ -129,4 +129,11 @@ public enum Constants {
     public static let enabled = "enabled"
     public static let agreement = "agreement"
     public static let useDynamicAgreement = "useDynamicAgreement"
+
+    // MARK: - SDK Integrator
+    public static let METADATA = "METADATA"
+    public static let payload = "payload"
+    public static let sdkMetadata = "sdkMetadata"
```
george-bafaloukas-forgerock: Heads up — `sdkMetadata` is added here but doesn't seem to be referenced anywhere. `MetadataCollector.key` is populated dynamically from the server's `key` field rather than pinned to this constant, so right now it's a bit of dead decoration, and if the server ever renames the field this constant would silently drift out of sync. Might be worth either wiring it in as the actual source of truth or dropping it.

Davinci/Davinci.xcodeproj/project.pbxproj
```diff
@@ -570,6 +577,7 @@
 				AA1A00042F0000000000AA01 /* PollingCollectorTests.swift in Sources */,
 				AA1A00082F0000000000AA01 /* QRCodeCollectorTests.swift in Sources */,
 				BB4928032F0900000000BB01 /* ReadOnlyTextCollectorTests.swift in Sources */,
+			BB4928072F0900000000BB01 /* MetadataCollectorTests.swift in Sources */,
```
george-bafaloukas-forgerock: Tiny nit, purely cosmetic: this line uses 3 tabs of indentation where every sibling entry in this block uses 4. Doesn't affect the build at all, just flagging for consistency if you're touching this file again.

george-bafaloukas-forgerock (commented):
Nice work on this one — the pause/resume model is clean and the parity with the Android/JS contract is spot on. A couple of things caught my eye while going through it, added as inline notes below. One more that doesn't have a line to anchor to: it looks like `CHANGELOG.md` wasn't updated — every other recent DaVinci collector (Polling, QrCode, SingleCheckbox, ReadOnlyText) got a one-liner in the `[2.1.0]` section, so it'd be good to add `MetadataCollector` there too for consistency. 🙂

DavinciPlugin/DavinciPlugin/Constants.swift
```diff
@@ -129,4 +129,11 @@ public enum Constants {
...
+    public static let action = "action"
```
spetrov, line 136: `Constants.ACTION = "ACTION"` already exists (line 47). Is the letter case significant? - i.e. can we use the existing one, or perhaps `Constants.ACTION.lowercased()` if the usage is case-sensitive?

spetrov (approved):
LGTM. Only one minor question. . .

Davinci/Davinci/collector/MetadataCollector.swift
```diff
@@ -0,0 +1,143 @@
...
+    ///   - code: A short error code string (e.g. `"USER_CANCELLED"`).
+    ///   - message: A human-readable description of the error.
+    ///   - isClientError: When `true`, tells the connector this originated
+    ///     on the client. Omit to leave the field out of the payload.
+    public func setError(code: String, message: String, isClientError: Bool? = nil) {
+        var error: [String: Any] = [
+            Constants.code: code,
+            Constants.message: message
+        ]
+        if let isClientError = isClientError {
```
vibhorgoswami: As discussed, Android is also not using this anymore. We can remove it from iOS.

Davinci/Davinci/collector/MetadataCollector.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// ```json
+    /// { "error": { "code": "...", "message": "...", "isClientError": true } }
+    /// ```
+    /// The server-side connector reads this and routes the flow to the
+    /// false/error branch.
+    ///
+    /// - Parameters:
+    ///   - code: A short error code string (e.g. `"USER_CANCELLED"`).
+    ///   - message: A human-readable description of the error.
+    ///   - isClientError: When `true`, tells the connector this originated
```
vibhorgoswami: Removing the comments as well.

vibhorgoswami (approved):
LGTM! Just 2 comments to have parity with Android.

Davinci/Davinci/collector/MetadataCollector.swift
vahancouver: Good catch. Fixed in this PR — `shouldValidate(node:)` in the sample app now casts to `any Validator`  instead of `ValidatedCollector`, so `MetadataCollector.validate()` is properly invoked. Also added a -  Note: to the `validate()` doc comment making it explicit that the SDK does not call this  automatically and the app is responsible for checking it before continuing the flow.

Davinci/Davinci/collector/MetadataCollector.swift
vahancouver: Addressed alongside the `validate()` fix above — the doc comment now explicitly states that validation is not enforced by the SDK and is the app's responsibility. No conformance change needed.

Davinci/Davinci/collector/MetadataCollector.swift
vahancouver: Fair point. `MetadataCollector` doesn't hold a DaVinci reference (unlike PollingCollector which has one wired post-init), so we can't route through the flow logger here without a bigger restructure.  The silent default to [:] is consistent with the existing SDK pattern for malformed JSON across  other collectors. Happy to revisit if you feel strongly, but leaving as-is for now.

DavinciPlugin/DavinciPlugin/Constants.swift
vahancouver: Removed. The key is server-driven so there's no value in pinning it client-side.

Davinci/Davinci.xcodeproj/project.pbxproj
vahancouver: Fixed — corrected to 4 tabs to match sibling entries.

DavinciPlugin/DavinciPlugin/Constants.swift
vahancouver: They're distinct values serving different purposes — `Constants.action = "action"` (lowercase) is the eventType sent in the resume POST body, which is what the DaVinci protocol expects on the wire. `Constants.ACTION = "ACTION"` (uppercase) is the server-sent collector type string used for factory dispatch. Case is significant and they can't be shared. 
I know Constants.ACTION.lowercased() can be used but leaving it as it is as they serve different purposes.

Davinci/Davinci/collector/MetadataCollector.swift
vahancouver: Done — isClientError parameter removed from setError, matching Android. Updated all call sites  (sample app, tests) and removed the now-unused Constants.isClientError constant.

Davinci/Davinci/collector/MetadataCollector.swift
vahancouver: Done — isClientError parameter removed from setError, matching Android. Updated all call sites (sample app, tests) and removed the now-unused Constants.isClientError constant.

witrisna (approved):

george-bafaloukas-forgerock (approved):
LGTM, just resolve the conflicts first

DavinciPlugin/DavinciPlugin/Constants.swift
spetrov: Thanks, sound good then.

spetrov (approved):
LGTM!

PR #180, opened 2026-07-23
Enhance FIDO2 Error Handling

vahancouver:
[SDKS-4478](https://pingidentity.atlassian.net/browse/SDKS-4478) [iOS] Enhance FIDO2 Error Handling

## Summary by CodeRabbit

- **New Features**
  - Exposed standardized FIDO failure context via an action key and added a new `"action"` event type when errors occur.
  - Collectors’ JSON output now prefers action keys when available, otherwise falls back to form data.

- **Bug Fixes**
  - Improved collector cleanup: authentication/registration collectors now clear recorded error state (and related payload signals) on reuse.
  - Sample app flows now continue correctly after failed FIDO authentication/registration.

- **Tests**
  - Expanded coverage for error propagation, event type switching, payload behavior, action key propagation, JSON output, and close/reuse behavior.

### DIFF
--- a/Davinci/Davinci/collector/Collectors.swift
+++ b/Davinci/Davinci/collector/Collectors.swift
@@ -18,8 +18,22 @@ public typealias Collectors = [any Collector]
 extension Collectors {
     /// Finds the event type from a list of collectors.
     /// This function iterates over the list of collectors and returns the value if the collector's value is not empty.
+    /// Submit/Flow collectors take precedence over ActionKeyProvider collectors so a failed FIDO
+    /// collector cannot shadow an explicit user action in the same node.
     /// - Returns: The event type as a String if found, otherwise nil.
     public func eventType() -> String? {
+        // First pass: honour explicit Submit/Flow actions.
+        for collector in self {
+            switch collector {
+            case let collector as SubmitCollector where !collector.value.isEmpty:
+                return collector.eventType()
+            case let collector as FlowCollector where !collector.value.isEmpty:
+                return collector.eventType()
+            default:
+                break
+            }
+        }
+        // Second pass: fall back to ActionKeyProvider-based event types (e.g. FIDO errors).
         for collector in self {
             if let submittable = collector as? Submittable {
                 if collector.payload() != nil {
@@ -33,6 +47,8 @@ extension Collectors {
     /// Represents a list of collectors as a JSON object for posting to the server.
     /// This function takes a list of collectors and represents it as a JSON object. It iterates over the list of collectors,
     /// adding each collector's key and value to the JSON object if the collector's value is not empty.
+    /// `SubmitCollector` and `FlowCollector` always take precedence for `actionKey`; an `ActionKeyProvider`
+    /// (e.g. a failed FIDO collector) only sets `actionKey` when it has not already been set.
     /// - Returns: JSON object representing the list of collectors.
     public func asJson() -> [String: Any] {
         var jsonObject: [String: Any] = [:]
@@ -49,7 +65,8 @@ extension Collectors {
                 }
             default:
                 if let actionKeyProvider = collector as? any ActionKeyProvider,
-                   let key = actionKeyProvider.actionKey {
+                   let key = actionKeyProvider.actionKey,
+                   jsonObject[Constants.actionKey] == nil {
                     jsonObject[Constants.actionKey] = key
                 } else if let fieldCollector = collector as? (any AnyFieldCollector),
                           let payload = fieldCollector.anyPayload() {
--- a/Davinci/Davinci/collector/Collectors.swift
+++ b/Davinci/Davinci/collector/Collectors.swift
@@ -2,7 +2,7 @@
 //  Collectors.swift
 //  PingDavinciPlugin
 //
-//  Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -48,7 +48,11 @@ extension Collectors {
                     jsonObject[Constants.actionKey] = collector.id
                 }
             default:
-                if let fieldCollector = collector as? (any AnyFieldCollector), let payload = fieldCollector.anyPayload() {
+                if let actionKeyProvider = collector as? any ActionKeyProvider,
+                   let key = actionKeyProvider.actionKey {
+                    jsonObject[Constants.actionKey] = key
+                } else if let fieldCollector = collector as? (any AnyFieldCollector),
+                          let payload = fieldCollector.anyPayload() {
                     formData[fieldCollector.id] = payload
                 }
             }
--- a/Fido/Fido/Davinci/AbstractFidoCollector.swift
+++ b/Fido/Fido/Davinci/AbstractFidoCollector.swift
@@ -15,12 +15,18 @@ import PingOrchestrate
 import AuthenticationServices
 
 /// An abstract base class for Fido collectors in a DaVinci flow.
-public class AbstractFidoCollector: AnyFieldCollector, DaVinciAware, Submittable, @unchecked Sendable {
-    
+public class AbstractFidoCollector: AnyFieldCollector, DaVinciAware, Submittable, ActionKeyProvider, @unchecked Sendable {
+
     public private(set) var type: String = ""
     public private(set) var key: String = ""
     public private(set) var label: String = ""
     public private(set) var required: Bool = false
+
+    /// DOMException name set when a FIDO operation fails; drives the DaVinci error response.
+    public var errorCode: String?
+
+    /// Supplies the DOMException name as `actionKey` when a FIDO error has occurred.
+    public var actionKey: String? { errorCode }
     
     /// The UUID of the field collector.
     public var id: String {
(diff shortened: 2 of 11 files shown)


Davinci/Davinci/collector/Collectors.swift
```diff
@@ -48,7 +48,11 @@ extension Collectors {
                     jsonObject[Constants.actionKey] = collector.id
                 }
             default:
-                if let fieldCollector = collector as? (any AnyFieldCollector), let payload = fieldCollector.anyPayload() {
+                if let actionKeyProvider = collector as? any ActionKeyProvider,
```
witrisna, line 67: If it is an ActionKeyProvider and with payload(), the payload() will be ignored?

vibhorgoswami (approved):
LGTM!

Fido/Fido/Davinci/AbstractFidoCollector.swift
```diff
@@ -94,64 +100,80 @@ public class AbstractFidoCollector: AnyFieldCollector, DaVinciAware, Submittable
...
-        
+
         switch nsError.domain {
         case ASAuthorizationError.errorDomain:
             switch nsError.code {
             case ASAuthorizationError.canceled.rawValue:
                 logger.d("Credential operation cancelled")
+                errorCode = FidoConstants.ERROR_NOT_ALLOWED
+                return .unsupportedAction(FidoConstants.ERROR_NOT_ALLOWED_MESSAGE)
+            case ASAuthorizationError.failed.rawValue:
```
witrisna, line 153: Does iOS align the spec? why convert failed.rawValue to NOT_ALLOWED, webauthn defined error code, however, not sure if iOS align with it.

Fido/Fido/Davinci/FidoAuthenticationCollector.swift
```diff
@@ -49,8 +50,13 @@ public class FidoAuthenticationCollector: AbstractFidoCollector, Closeable, @unc
     
     /// The payload to be sent to the DaVinci server.
     ///
-    /// - Returns: A dictionary containing the assertion value, or `nil` if authentication has not been completed.
+    /// Returns an empty dictionary when a FIDO error has occurred (non-nil sentinel so
+    /// `Collectors.eventType()` picks up this collector and the `actionKey` error path fires).
+    /// Returns `nil` when no FIDO operation has been completed yet.
     override public func payload() -> [String: Any]? {
+        if errorCode != nil {
```
witrisna, line 57: Do we need errorCode null check? if failed assertionValue should be null and it should return null for payload().

Fido/Fido/Davinci/FidoRegistrationCollector.swift
```diff
@@ -48,8 +49,13 @@ public class FidoRegistrationCollector: AbstractFidoCollector, Closeable, @unche
     
     /// The payload to be sent to the DaVinci server.
     ///
-    /// - Returns: A dictionary containing the attestation value, or `nil` if registration has not been completed.
+    /// Returns an empty dictionary when a FIDO error has occurred (non-nil sentinel so
+    /// `Collectors.eventType()` picks up this collector and the `actionKey` error path fires).
+    /// Returns `nil` when no FIDO operation has been completed yet.
     override public func payload() -> [String: Any]? {
+        if errorCode != nil {
```
witrisna, line 56: Do we need errorCode null check? if failed assertionValue should be null and it should return null for payload().

Fido/Fido/Davinci/AbstractFidoCollector.swift
vahancouver: The WebAuthn spec (§10.2 Web Authentication API (https://www.w3.org/TR/webauthn-2/#sctn-automation-getAssertion)) says a user-cancelled or user-denied biometric prompt should surface as NotAllowedError. On iOS, ASAuthorizationError.canceled (.rawValue = 1001) maps to user cancellation, and ASAuthorizationError.failed (.rawValue = 1004) covers biometric/device auth failure — both are user-agent-denied scenarios. The WebAuthn spec uses NotAllowedError for both. This is the same mapping used by browsers implementing WebAuthn on iOS (WKWebView). So yes, iOS does align with the spec for these two codes.

Fido/Fido/Davinci/FidoAuthenticationCollector.swift
vahancouver: After a failure, assertionValue (or attestationValue) is indeed nil — so without the errorCode check, payload() would fall through to return nil. But a nil payload means Collectors.eventType() skips this collector entirely, and the DaVinci flow would never send the error's actionKey to the server.

Fido/Fido/Davinci/FidoRegistrationCollector.swift
vahancouver: After a failure, assertionValue (or attestationValue) is indeed nil — so without the errorCode check, payload() would fall through to return nil. But a nil payload means Collectors.eventType() skips this collector entirely, and the DaVinci flow would never send the error's actionKey to the server.

Fido/Fido/FidoConstants.swift
```diff
@@ -158,9 +160,7 @@ public struct FidoConstants {
     public static let WEB_AUTHN_OUTCOME = "webAuthnOutcome"
 
     // MARK: - Error Types
-    
-    /// An unsupported error type.
-    public static let ERROR_UNSUPPORTED = "unsupported"
+
```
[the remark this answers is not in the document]
vahancouver: Re-added ERROR_UNSUPPORTED = "unsupported" as a @available(*, deprecated, renamed: "ERROR_NOT_SUPPORTED") constant to avoid a breaking change for existing callers.

Fido/Fido/Davinci/FidoAuthenticationCollector.swift
george-bafaloukas-forgerock, line 57: `errorCode` is cleared in `close()` (good), but it's never cleared on a successful retry within the same collector instance. If `authenticate()` fails once (setting `errorCode`) and then succeeds on a subsequent call without an intervening `close()`, `assertionValue` gets set correctly, but this check still sees the stale `errorCode` and returns `[:]` — silently dropping the real success payload instead of sending the assertion.

Could we clear `errorCode` at the top of the success path in `authenticate()` (around where `assertionValue` is assigned), similar to how `close()` resets it? Same issue applies to `FidoRegistrationCollector.register()`.

Fido/Fido/Davinci/FidoRegistrationCollector.swift
george-bafaloukas-forgerock, line 56: Same stale-`errorCode`-on-retry issue as `FidoAuthenticationCollector.payload()` — a successful `register()` call after a prior failure (without an intervening `close()`) will still hit this branch and return `[:]` instead of the real attestation payload, because `errorCode` from the earlier failure is never cleared on success.

Davinci/Davinci/collector/Collectors.swift
george-bafaloukas-forgerock, line 67: Picking up witrisna's comment on this line from the earlier review round, which doesn't look like it got a reply yet.

With this change, a Fido collector that has failed (but not yet been `close()`'d) always has a non-nil `payload()` (`[:]`) and a non-nil `actionKey`. If that collector is co-located in the same node as a `FlowCollector`/`SubmitCollector` (a real pattern — see `ContinueNodeView.swift`), this loop will happily overwrite `jsonObject[Constants.actionKey]` with the stale FIDO error's key after the user's actual Cancel/Submit action was already written, depending on array order. `eventType()` above has the same first-match hazard for the same reason.

Could we scope this so a failed-but-not-closed Fido collector can't clobber another collector's action in the same node — e.g. only apply the `ActionKeyProvider` branch when no other `Submittable` in the list has already produced a payload, or make the ordering/precedence explicit? Would also be good to close the loop with witrisna on this thread either way.

george-bafaloukas-forgerock (changes requested):
Thanks for adding this — the DOMException-to-`actionKey` mapping approach is a nice, minimal way to surface FIDO errors to the DaVinci server, and the `close()`/state-reset wiring for the happy path looks solid. I found two issues around collector state that I think are worth resolving before merge, left as inline comments. Happy to take another pass once addressed!

witrisna (approved):

Fido/Fido/Davinci/FidoAuthenticationCollector.swift
vahancouver: Added self.errorCode = nil on the success path (just before storing assertionValue/attestationValue). This ensures a successful retry after a prior failure doesn't leave stale error state that causes payload() to return [:] and drop the real result.

Fido/Fido/Davinci/FidoRegistrationCollector.swift
vahancouver: Added self.errorCode = nil on the success path (just before storing assertionValue/attestationValue). This ensures a successful retry after a prior failure doesn't leave stale error state that causes payload() to return [:] and drop the real result.

Davinci/Davinci/collector/Collectors.swift
vahancouver: I didn't address Andy comment as he marked it resolved. 
To address the concerns about FlowCollector/SubmitCollector eventType/ActionKey being overrided, here are the changes I made.
1. In eventType(): Added a first pass that explicitly checks for SubmitCollector/FlowCollector with non-empty values before falling through to the generic Submittable sweep. This gives explicit user actions priority over FIDO error event types regardless of array order.
2. In asJson() : Added a jsonObject[Constants.actionKey] == nil guard on the ActionKeyProvider branch. A failed FIDO collector processed before a SubmitCollector/FlowCollector in the same node can no longer overwrite the user's actual cancel/submit action.

george-bafaloukas-forgerock (approved):
LGTM

spetrov (approved):
LGTM! 👍🏻

PR #181, opened 2026-07-24
P1RECMOB-3476: Add recognize module and journey integration

gabrielocci:
**JIRA Ticket:** [P1RECMOB-3030](**JIRA Ticket:** [P1RECMOB-3476](https://pingidentity.atlassian.net/browse/P1RECMOB-3476)

  ## Summary

  Introduces the `PingRecognize` module — a wrapper around the Keyless SDK (v6.0.0) that bridges PingAM Journey flows
  with biometric recognition (face enroll/authenticate).

  **What's in scope:**

  - **`Recognize/` module**: `RecognizeError` (with error code propagation), `RecognizeCallback`, and
  `RecognizeCallbacks` registry registration
  - **Journey integration ✅**: `RecognizeCallback` handles the full setup → enroll/authenticate lifecycle driven by
  `AbstractCallback` and `ContinueNodeAware`; operation type (`ENROLL` / `AUTHENTICATE`) is server-driven via the
  `operationType` output field
  - **`enrollWithClientState` support**: when the server supplies a `clientState` alongside `operationType =
  AUTHENTICATE`, the callback transparently calls `validateUserDeviceActive` and falls back to enrollment if the user is
  not yet enrolled
  - **Error propagation**: Keyless SDK error codes forwarded to the server via `IDToken1clientError` /
  `IDToken1clientErrorCode`
  - **Sample app wired** to a Recognize Journey for manual testing (`RecognizeCallbackView`)
  - **Unit tests** covering `RecognizeError`, `RecognizeMobileSDKOptions`, `RecognizeCallback` output field parsing,
  input setters, and operation type constants (493 lines)
  - **SPM and CocoaPods** distribution (`PingRecognize.podspec`, `Package.swift`)

  **What's NOT in scope:** DaVinci integration (deferred to next iteration).

  ---

  ## Review guidance

  - Start here: `Recognize/Recognize/Journey/RecognizeCallback.swift` — the correctness-critical code; covers configure
  → enroll/authenticate/enrollWithClientState lifecycle
  - `Recognize/Recognize/Journey/RecognizeCallbacks.swift` — callback registry registration via `NSClassFromString`
  (mirrors the pattern in `Journey.swift`)
  - `Recognize/RecognizeTests/RecognizeTests.swift` — primary test coverage
  - `SampleApps/PingExample/PingExample/Callbacks/RecognizeCallbackView.swift` — sample app integration
  - Error handling: `RecognizeError` wraps `KeylessSDKError` codes — worth checking propagation is complete through
  `execute()` → `error()` → `setClientErrorCode()`

  ---

  ## ⚠️  Build prerequisite — Keyless SDK registry access

  The Keyless SDK is fetched from a private Swift Package Registry. `Package.swift` declares:

  ```swift
  .package(id: "keyless.mobile-sdk", from: "6.0.0")

  For local builds, the registry must be configured in Xcode under Settings → Accounts → Swift Package Registry with
  valid Keyless credentials. Without this, SPM resolution will fail before any module compiles.

  For CI, the registry credentials must be injected via Xcode Cloud / GitHub Actions secrets.

  Credentials should be obtainable from the Keyless/partners team or your internal secrets store.
  ```
)

## Summary by CodeRabbit

* **New Features**
  * Added PingOne Recognize biometric enrollment and authentication for iOS.
  * Added device validation, client-state restoration, selfie capture, configurable SDK options, and detailed operation results.
  * Added Swift Package Manager and CocoaPods distribution support.
  * Integrated Recognize flows into the sample application and Journey experience.

* **Documentation**
  * Added installation, integration, usage, troubleshooting, and licensing guidance.

* **Tests**
  * Added comprehensive coverage for Recognize operations, callbacks, validation, errors, and configuration.


Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+/// enrollment and authentication. The `operationType` output field distinguishes the two flows.
+///
+/// **Usage:**
+/// ```swift
+/// if let recognizeCallback = node.callbacks.first(where: { $0 is RecognizeCallback })
+///         as? RecognizeCallback {
+///     let result = await recognizeCallback.execute()
+/// }
+/// ```
+public class RecognizeCallback: AbstractCallback, ContinueNodeAware, @unchecked Sendable {
```
george-bafaloukas-forgerock: We would need to align iOS and Android to have the same class names and signatures where possible. We aim for a full alignment par specific platform eventualities. 
My preference would be to stick to the pattern of splitting the callbacks and having a `PingOneRecognizeEnrollCallback` and a `PingOneRecognizeAuthenticateCallback` similar to other modules in the SDK and how the Recognize Android module has been structured

george-bafaloukas-forgerock: This would also help split this file to smaller ones

george-bafaloukas-forgerock: **Suggest splitting this into an abstract base + two concrete callbacks.**

Following up on my earlier comment — having now checked this against the rest of the SDK, the split matches both existing iOS patterns:

- **Protect**: `AbstractProtectCallback` (shared parsing/input helpers) + `PingOneProtectInitializeCallback` / `PingOneProtectEvaluationCallback`
- **Fido**: `FidoCallback` (shared base) + `FidoRegistrationCallback` / `FidoAuthenticationCallback`

So `AbstractRecognizeCallback` + `PingOneRecognizeEnrollCallback` + `PingOneRecognizeAuthenticateCallback` would be right at home, and matches the Android naming exactly.

Alongside the file-size win, the split gives you a shared config builder, which fixes the `enrollWithClientState` divergence noted further down — the two `BiomEnrollConfig` constructions stop being able to drift apart.

**One open question worth deciding together.** In Protect and Fido, the enroll-vs-auth selection happens *externally*, in `MetadataCallback.initialize(with:)` (`Journey/Journey/Callbacks/MetadataCallback.swift:42-70`), which looks at the metadata and instantiates the right concrete type. Since the agreed contract makes `PingOneRecognizeCallback` a first-class typed callback, there's no metadata dispatcher to hang that off, so the selection needs a home. Two options:

1. **A factory callback** — register `RecognizeCallback` for `"PingOneRecognizeCallback"`, have its `initialize(with:)` read `operationType` and return the concrete instance. This is Android's shape, and `CallbackRegistry` does honour the return value (`JourneyPlugin/JourneyPlugin/CallbackRegistry.swift:57`), so it works. Worth noting it'd be the first callback on iOS to specialise itself this way — `ReCaptchaEnterpriseCallback.initialize(with:)` overrides only to scan `input[]` and always returns `self`.
2. **Register all three type strings** and let each concrete ignore a mismatched `operationType`.

I'd lean towards (1) for Android parity, but you're closer to the server contract — your call.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+    /// Input field key for the client state produced by the Keyless SDK.
+    public static let inputClientState = "IDToken1clientState"
+    /// Input field key for the Recognize user ID produced during enrollment.
+    public static let inputRecognizeId = "IDToken1recognizeId"
+    /// Input field key for the device public signing key retrieved after authentication.
+    public static let inputDevicePublicSigningKey = "IDToken1devicePublicSigningKey"
+    /// Input field key for the client error message.
+    public static let inputClientError = "IDToken1clientError"
+    /// Input field key for a structured client error code.
+    public static let inputClientErrorCode = "IDToken1clientErrorCode"
```
george-bafaloukas-forgerock: **Input field names: consider suffix matching rather than a hardcoded `IDToken1` prefix.**

These six constants pin the prefix to `IDToken1`, and `input(_:forKey:)` matches on exact equality (`JourneyPlugin/JourneyPlugin/AbstractCallback.swift:72-76`). If the server ever emits `IDToken2signedJwt` — which happens when the callback isn't first in the node — `firstIndex(where:)` finds nothing, the function returns `json` untouched, and every write silently no-ops. The Journey would then submit an empty JWT with no error surfaced.

No other iOS module hardcodes an `IDToken<N>` prefix. `PingBinding` matches on suffix instead:

```swift
// Binding/Binding/Callback/DeviceBindingCallback.swift:157
if let index = inputArray.firstIndex(where: { ($0[JourneyConstants.name] as? String)?.hasSuffix(name) ?? false }) {
```

`ReCaptchaEnterpriseCallback` takes a similar approach, discovering the actual input names at `initialize(with:)` time via `contains`. Android also keys off suffixes (`SIGNED_JWT_SUFFIX`, `CLIENT_STATE_SUFFIX`, …), so suffix matching aligns both platforms here.

Suggest storing the suffixes (`"signedJwt"`, `"clientState"`, …) and matching with `hasSuffix`.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+}
+
+// MARK: - JourneyConstants (Recognize output field keys + defaults)
+
+extension JourneyConstants {
+    /// Fallback error string used when the error localised description is empty.
+    public static let clientError = "clientError"
```
george-bafaloukas-forgerock: **`clientError` here collides with the one `PingProtect` already declares.**

`Protect/Protect/Journey/AbstractProtectCallback.swift:104` already adds `public static let clientError = "clientError"` to this same `public enum JourneyConstants`. Two modules adding an identically-named public static member to the same extended enum makes the reference ambiguous for any app that imports both:

```
error: ambiguous use of 'clientError'
note: found this candidate in module 'PingProtect'
note: found this candidate in module 'PingRecognize'
```

`SampleApps/PingExample/PingExample/JourneyView.swift` already imports both modules — it only escapes today because it never references the constant.

The good news is that `extension JourneyConstants` is very much the house style (Protect, ReCaptcha and DeviceProfile all do it), so nothing structural needs to change. `PingReCaptchaEnterprise` already shows the fix — it declares the same key without `public`:

```swift
// ReCaptchaEnterprise/ReCaptchaEnterprise/ReCaptchaEnterprise.swift:44
static let clientError = "clientError"
```

Dropping `public` on `clientError` resolves it. I diffed all 36 constants this extension adds against the base `JourneyConstants` — no other overlaps, so this is the only one.

Two small extras while you're here: `boolTrue`/`boolFalse` are only referenced from tests, and `webSDKOptions` isn't referenced at all — probably worth keeping both out of the public surface, since public API is harder to walk back later.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+    /// Values that are not already strings are coerced via string interpolation.
+    private static func stringDictionary(from value: Any) -> [String: String] {
+        if let dict = value as? [String: String] { return dict }
+        if let dict = value as? [String: Any] {
+            return dict.reduce(into: [:]) { result, pair in
+                result[pair.key] = "\(pair.value)"
+            }
+        }
+        return [:]
+    }
```
george-bafaloukas-forgerock: **`mobileSDKOptions` booleans are silently dropped when the server sends real JSON booleans.**

Non-string values get coerced with `"\(pair.value)"`. Values decoded by `JSONSerialization` come through as `__NSCFBoolean`, which interpolates to `"1"` / `"0"` rather than `"true"` / `"false"`, and `Bool.init(String)` returns `nil` for both:

```
f as? Bool = false | interp = "0" | Bool(interp) = nil
t as? Bool = true  | interp = "1" | Bool(interp) = nil
```

So if the server sends:

```json
"mobileSDKOptions": { "showInstructionsScreen": false, "showSuccessFeedback": false, "livenessEnvironmentAware": false }
```

all three computed properties return `nil` and `enroll()` falls back to `BiomEnrollConfig.DEFAULT_*` — the instructions screen and success overlay still appear despite being disabled server-side.

`initValue` already handles this correctly for the top-level `generateClientState` field (line 173 checks `value as? Bool` first) — it's just the nested dictionary path that misses it. `Int` coercion is unaffected (`"4"` → `4`), which is why the existing tests pass.

Suggest checking `as? Bool` (and `as? NSNumber`) before falling back to interpolation.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+            livenessConfiguration: Self.livenessConfiguration(from: options.livenessConfiguration),
+            // BiomEnrollConfig has no DEFAULT_LIVENESS_ENV_AWARE; BiomAuthConfig's constant carries the same value.
+            livenessEnvironmentAware: options.livenessEnvironmentAware ?? BiomAuthConfig.DEFAULT_LIVENESS_ENV_AWARE,
+            cameraDelaySeconds: options.cameraDelaySeconds ?? BiomEnrollConfig.DEFAULT_DELAY,
+            generatingClientState: generateClientState ? .backup : nil,
+            showInstructionsScreen: options.showInstructionsScreen ?? BiomEnrollConfig.DEFAULT_SHOW_INSTRUCTIONS_SCREEN,
+            showSuccessFeedback: options.showSuccessFeedback ?? BiomEnrollConfig.DEFAULT_SHOW_SUCCESS_FEEDBACK,
+            showFailureFeedback: options.showFailureFeedback ?? BiomEnrollConfig.DEFAULT_SHOW_FAILURE_FEEDBACK,
+            presentationStyle: Self.enrollPresentationStyle(from: options.presentation)
+        )
```
george-bafaloukas-forgerock: **This `BiomEnrollConfig` is missing `operationInfo` and `jwtSigningInfo`.**

Compared with the one built in `enroll()` (lines 289-302), this construction omits both `operationInfo:` and `jwtSigningInfo:`.

Concretely: server sends `operationType = AUTHENTICATE` with a `clientState`, non-empty `transactionData` and an `operationInfoId`, and the user isn't yet enrolled. The enrollment then proceeds with no signed transaction claim and no binding to the server's operation ID, so the JWT submitted back in the signed-JWT input omits the transaction data the server expects to verify.

`Recognize/README.md` documents `transactionData → jwtSigningInfo` as unconditional for enrollment, so the docs and behaviour disagree too.

Android avoids this by having a single `buildEnrollConfig(clientStateOverride:)` builder shared by both paths. Extracting a shared builder here (which falls out naturally from the enroll/authenticate split) would prevent the two from drifting again.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+    private(set) public var mobileSDKOptions: RecognizeMobileSDKOptions = RecognizeMobileSDKOptions(raw: [:])
+
+    // MARK: - Initialization
+
+    /// Populates output properties from the JSON output fields delivered by the server.
+    public override func initValue(name: String, value: Any) {
+        switch name {
+        case JourneyConstants.operationType:
+            if let stringValue = value as? String {
+                self.operationType = RecognizeOperationType(rawValue: stringValue) ?? .enroll
```
george-bafaloukas-forgerock: **Unknown or differently-cased `operationType` silently performs an enrollment.**

`RecognizeOperationType(rawValue: stringValue) ?? .enroll` means a server emitting `"authenticate"` (lowercase) or `"AUTHENTICATION"` would run the *enrollment* ceremony — re-enrolling a biometric for a user who was meant to be verified. Since this decides which authentication ceremony runs, failing closed feels safer than defaulting.

Worth noting every other server-supplied string in this file is normalised with `.uppercased()` first (lines 433, 445, 455), so `operationType` is the odd one out.

For comparison, `DeviceBindingCallback` does use a `?? ` fallback, but to a *neutral* value:

```swift
// Binding/Binding/Callback/DeviceBindingCallback.swift:57
deviceBindingAuthenticationType = DeviceBindingAuthenticationType(rawValue: authType) ?? .none
```

`.none` doesn't pick a ceremony; `.enroll` does. Android throws `IllegalArgumentException` for an unrecognised value.

Suggest at minimum `.uppercased()` before the lookup, and ideally surfacing an error for unsupported values rather than defaulting.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+
+    // MARK: - Execution
+
+    /// Configures the Keyless SDK and executes the enrollment or authentication operation.
+    ///
+    /// - Returns: `.success(())` on completion, or `.failure(error)` if any step fails.
+    ///   On failure the `IDToken1clientError` input field is automatically populated.
+    public func execute() async -> Result<Void, Error> {
```
george-bafaloukas-forgerock: **Consider returning the operation result rather than `Void`.**

`execute()` returns `Result<Void, Error>`, so callers can't get at the signed JWT, client state, keylessId, or a captured selfie frame — the values are written into the input fields and then discarded.

Every other callback in the SDK returns something:

| Module | Signature |
|---|---|
| Fido | `Result<[String: Any], Error>` |
| Binding | `Result<[String: Any], Error>` |
| Protect (evaluation) | `Result<String, Error>` |
| ReCaptcha | `Result<String, Error>` |
| ExternalIdP | `Result<IdpResult, IdpExceptions>` |

The only `Result<Void, Error>` is `PingOneProtectInitializeCallback.start()`, which genuinely produces no value.

`IdpResult` (`ExternalIdP/ExternalIdP/IdpResult.swift`) is the precedent for a small dedicated struct, so something like:

```swift
public struct RecognizeResult: Sendable {
    public let signedJwt: String?
    public let clientState: String?
    public let recognizeId: String
    public let selfie: UIImage?   // or CGImage
}
```

Android returns the equivalent `RecognizeSuccess` from both paths. I'd suggest `RecognizeResult` for the iOS name to match the `IdpResult` convention.

Two related gaps while you're in here:

- **`retrieveSelfie`** — Android exposes `enroll(retrieveSelfie:)` / `authenticate(retrieveSelfie:)`, mapping to `shouldRetrieveEnrollmentFrame` / `shouldRetrieveAuthenticationFrame`. Neither is set on iOS, so the frame can never be returned.
- **`audience`** — the server field isn't parsed at all here. Android's `buildJwtSigningInfo()` adds it to `JwtSigningInfo` when non-blank, whereas `jwtSigningInfo(from:)` (line 256) passes only `claimTransactionData`, so the JWT goes out without its audience claim.

george-bafaloukas-forgerock (commented):
Thanks for putting this together — the module is well documented and the Keyless wrapping reads cleanly. I've gone through it alongside the Android PR ([#229](https://github.com/ForgeRock/ping-android-sdk/pull/229)) and checked each point against how `PingProtect`, `PingFido` and `PingBinding` do things, so the suggestions below are aimed at conformity with the rest of the iOS SDK rather than mirroring Android literally.


**Main themes:**

1. **Split the callback** into `AbstractRecognizeCallback` + `PingOneRecognizeEnrollCallback` + `PingOneRecognizeAuthenticateCallback`. This matches Protect (`AbstractProtectCallback` + `PingOneProtectInitializeCallback`/`PingOneProtectEvaluationCallback`) and Fido (`FidoCallback` + `FidoRegistrationCallback`/`FidoAuthenticationCallback`), aligns the class names with Android, and breaks up a 537-line file. There's one open design question about where `operationType` selection lives — details inline.
2. **Input field names** are hardcoded as `IDToken1…` and matched exactly. Suffix matching (as `PingBinding` does) would be more robust.
3. **A few correctness bugs** around `mobileSDKOptions` boolean parsing, the `enrollWithClientState` config, and the unknown-`operationType` fallback.
4. **`clientError` collides** with the one `PingProtect` already adds to `JourneyConstants` — a compile error for apps importing both.
5. **The public API returns no value**, so callers can't get the JWT, client state or keylessId back.

One note on the Android side, for the record: the `MetadataCallback` / `derivedCallback` handling in `AbstractRecognizeCallback.kt` looks like it may not be needed, since the agreed contract defines `PingOneRecognizeCallback` as a first-class typed callback. Worth raising on #229 separately — nothing to change here.

Happy to pair on the restructuring if that's useful.

Recognize/Recognize/Journey/RecognizeCallback.swift
```diff
@@ -0,0 +1,537 @@
...
+    ///
+    /// Falls back to `.LEVEL_1` for unrecognised values.
+    private static func livenessConfiguration(from string: String) -> Keyless.LivenessConfiguration {
+        switch string.uppercased() {
+        case "DEVELOPMENT": return .DEVELOPMENT
+        case "LEVEL_1":     return .LEVEL_1
+        case "LEVEL_2":     return .LEVEL_2
+        default:            return .LEVEL_1
+        }
+    }
```
george-bafaloukas-forgerock: **Liveness mapping: a hardcoded switch will need a code change for each new level.**

The three-case switch with a `.LEVEL_1` fallback means any new liveness level the Keyless SDK adds needs an SDK release here before it can be used, and an unrecognised value silently downgrades to `LEVEL_1` rather than the SDK's own default.

Android uses `LivenessConfiguration.valueOf(...)` inside `runCatching` and falls through to the SDK default on failure, so new values work without a change. If `Keyless.LivenessConfiguration` is `RawRepresentable`, `Keyless.LivenessConfiguration(rawValue: string.uppercased())` with a nil-fallback to the SDK default would give the same property here.

Same reasoning applies to `enrollPresentationStyle(from:)` and `authPresentationStyle(from:)` just below.

Recognize/Recognize/Journey/RecognizeCallbacks.swift
```diff
@@ -0,0 +1,32 @@
...
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+#if canImport(KeylessSDK)
+import Foundation
+import PingJourneyPlugin
+
+@objc
+class RecognizeCallbacks: NSObject {
```
george-bafaloukas-forgerock, line 16: **Keeping this type internal is right — it's the README that needs the tweak.**

`internal class …Callbacks: NSObject` is the majority pattern (`ProtectCallbacks`, `IdpCallbacks`, `ReCaptchaEnterprise` are all internal), and since `Journey.swift:129` registers this dynamically via `NSClassFromString`, integrators never need to touch it. So no change needed here.

The mismatch is that `Recognize/README.md` (lines 88 and 232) tells integrators to call:

```swift
await RecognizeCallbacks.registerCallbacks()
```

which won't compile — the type isn't visible outside the module, and `registerCallbacks()` is synchronous (it wraps its work in an unawaited `Task`), so the `await` is invalid too.

Two ways to square it:

- **Drop the section from the README** (my preference) — dynamic registration already covers it, matching how Protect and ExternalIdP document it.
- **Or make it public**, as `PingFido` does with `CallbackInitializer` — but note Fido is public *precisely because* its README documents manual registration, and `CallbackInitializer` would also be the better name for cross-platform parity with Android's `CallbackInitializer`.

Either way the `await` should go.

Recognize/RecognizeTests/RecognizeTests.swift
```diff
@@ -0,0 +1,493 @@
...
+        XCTAssertEqual(JourneyConstants.mobileSDKOptions, "mobileSDKOptions")
+    }
+
+}
+
+// MARK: - EnrollWithClientStateTests
+
+final class EnrollWithClientStateTests: XCTestCase {
+
+    func testClientStateAbsentRoutesToAuthenticateBranch() async {
```
george-bafaloukas-forgerock, line 454: **Coverage gap on the execution path.**

56 tests here, but none reference `execute()`, `configure()`, `enroll(`, `authenticate(` or `enrollWithClientState` — which is the code the PR description calls out as "the correctness-critical code" and flags for error-propagation review.

`testClientStateAbsentRoutesToAuthenticateBranch` (line 440) asserts only that `clientState == ""` and `operationType == .authenticate` after `initialize`; it never invokes the branch in its name, so a regression in the routing logic would still pass.

Worth adding, ideally behind an injectable Keyless seam:

- `configure()` failure → `RecognizeError` code reaches the client-error-code input
- `enroll` / `authenticate` failure → client error and error code both populated
- `validateUserDeviceActive` returning a non-`userNotEnrolled` error → propagated, no enrollment attempted
- the boolean-parsing case above, with real JSON booleans rather than pre-stringified `"true"` / `"false"`

For reference, Android's `RecognizeCallbackTest.kt` covers this path fairly thoroughly (~1176 lines), so there may be test cases worth porting across.

PingRecognize.podspec
```diff
@@ -0,0 +1,39 @@
...
+  s.static_framework = true
+
+  s.ios.deployment_target = '16.0'
+
+  base_dir = "Recognize/Recognize"
+  s.source_files = base_dir + '/**/*.swift', base_dir + '/**/*.h'
+  s.resource_bundles = {
+    'Recognize' => [base_dir + '/*.xcprivacy']
+  }
+
```
george-bafaloukas-forgerock, line 36: **A couple of repo-convention artifacts that look like they were missed:**

- `generate-docs.sh` — the `MODULES` array (lines 14-39) doesn't include `Recognize`, so the new module gets no API docs. The `Generate API Docs` check is currently red.
- Root `README.md` — the module tree doesn't list `PingRecognize`.
- `CHANGELOG.md` — no entry yet. The Android PR has one; the last comparable iOS module addition (`AuthMigration`) added both a CHANGELOG line and a `release-pods.sh` entry.
- `PingRecognize.podspec:36` — pins `PingJourneyPlugin` to `~> 2.0.0`, whereas every other module at 2.1.0 pins `~> 2.1.0` (checked against `PingFido`, `PingBinding`, `PingOneProtect`, `PingReCaptchaEnterprise`, `PingJourney`).
- `release-pods.sh` — `LIBRARIES` doesn't include `PingRecognize`, so the pod wouldn't publish. `AGENTS.md:18` notes both distribution paths need to stay in sync.
- `Recognize.xcodeproj/project.pbxproj:311,332` — `MARKETING_VERSION = 2.0.0` while the podspec says `2.1.0` and sibling modules are all `2.1.0`.

Also two small README accuracy items: the enrollment mapping table lists `mobileSDKOptions.numberOfEnrollmentCircuits → BiomEnrollConfig.numberOfEnrollmentCircuits`, but the code passes it to `SetupConfig` in `configure()` (line 265); and the SwiftUI examples use `for await node in journey.start()`, whereas `start()` returns a single `Node` (there's no `AsyncSequence` conformance in `Journey/` or `Orchestrate/`). The sample app's `RecognizeCallbackView.swift` has the correct single-`Node` pattern.

george-bafaloukas-forgerock (changes requested):

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thank you, this issue has been solved [here](https://github.com/gabrielocci/ping-ios-sdk/commit/6d48d078a4620746be1a4d9d68db11323851aef6)

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: thank you, addressed [here](https://github.com/gabrielocci/ping-ios-sdk/commit/9e349f4ee1739f6dc63b7ad613b0f6d37a992330)

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks for catching this! Resolved across two commits:

  - `webSDKOptions` lost its `public` modifier here: https://github.com/gabrielocci/ping-ios-sdk/commit/6d48d078a4620746be1a4d9d68db11323851aef6 (side effect of splitting `RecognizeCallback` into
  `AbstractRecognizeCallback` / `PingOneRecognizeEnrollCallback` / `PingOneRecognizeAuthenticateCallback`)
  - `clientError` was made internal, and `boolTrue` / `boolFalse` were removed entirely (along with the tests asserting their literal values), fixing the collision with `PingProtect`'s
  `AbstractProtectCallback`: https://github.com/gabrielocci/ping-ios-sdk/commit/a1a41c1be050ed6d4f96679a2ca5e2279d495f3a

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks George! Fixed here: https://github.com/gabrielepaolocci/ping-ios-sdk/commit/2b1f534b074710fc6b82775c67f9cb41447dcabd

 Let me know if this looks good to you.

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks George! Fixed here: https://github.com/gabrielepaolocci/ping-ios-sdk/commit/fb894f7d523475ec079fdf9b6d098677bc1cfb8f
  
Let me know if this works for you.

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks George! Fixed here:
  https://github.com/gabrielepaolocci/ping-ios-sdk/commit/e347ef9cd4b19f2057b7bfd2b390ee5294d2fa15

  Added `RecognizeResult` (`signedJwt`, `clientState`, `recognizeId`, `selfie: CGImage?`) matching the `IdpResult`
  convention, and both `enroll`/`enrollWithClientState` and `performAuthenticate` now return it instead of `Void`. Also
  closed the two related gaps: `retrieveSelfie` is now wired to
  `shouldRetrieveEnrollmentFrame`/`shouldRetrieveAuthenticationFrame`, and `audience` is parsed from the server payload
  and passed into `JwtSigningInfo(claimTransactionData:audience:)`

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Fixed here: https://github.com/gabrielepaolocci/ping-ios-sdk/commit/64a2d6eacd52eb2a426fb09a94ef78d4cc714994

thanks George

Recognize/RecognizeTests/RecognizeTests.swift
gabrielocci: Thanks George! Fixed here:
https://github.com/gabrielepaolocci/ping-ios-sdk/commit/fb894f7d523475ec079fdf9b6d098677bc1cfb8f

Added the injectable Keyless seam you suggested (testable subclasses overriding `configure()`/`enroll()`/`enrollWithClientState()`/`performAuthenticate()`/`validateUserDeviceActive()`) and covered the execution/error-propagation paths you flagged, plus a couple more around clientState routing.

Haven't ported the Android test suite over, happy to do that as a follow-up if you think it's worth it.

PingRecognize.podspec
gabrielocci: Thanks George for the catch, fixed here: https://github.com/gabrielepaolocci/ping-ios-sdk/commit/27ede2a293b1e74aba53367051aaafe15fccf747
 
 Let me know if this looks good to you.

Recognize/Recognize/Journey/RecognizeCallbacks.swift
gabrielocci: Thanks George! Went with dropping the section, matching how Protect/ExternalIdP document it:
https://github.com/gabrielepaolocci/ping-ios-sdk/commit/3c2f21ff30ad0b21e45263498f15c6900e39391c

Removed both `await RecognizeCallbacks.registerCallbacks()` occurrences from the README (the section is gone entirely), dynamic registration via `NSClassFromString` already covers it. Also fixed the remaining examples' casts, which were stale after the enroll/authenticate split: `callback as? RecognizeCallback` → `callback as? PingOneRecognizeEnrollCallback` / `PingOneRecognizeAuthenticateCallback`.

  Let me know if this looks good to you.

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks George! Fixed here: https://github.com/gabrielepaolocci/ping-ios sdk/commit/fb894f7d523475ec079fdf9b6d098677bc1cfb8f

Slightly different shape than Android's `buildEnrollConfig(clientStateOverride:)`, but same effect: `enrollWithClientState` no longer builds its own `BiomEnrollConfig`, it delegates straight to `enroll(options:)`, which already includes both `operationInfo` and `jwtSigningInfo`

So the missing-claim/missing-binding gap is closed, and there's only one construction site left to drift.

Let me know if it solves the comment for you

Recognize/Recognize/Journey/RecognizeCallback.swift
gabrielocci: Thanks George! Went with the split you preferred, plus followed through on signature/naming alignment across three commits:

  - https://github.com/gabrielepaolocci/ping-ios-sdk/commit/6d48d078a4620746be1a4d9d68db11323851aef6 — split into
  `PingOneRecognizeEnrollCallback` / `PingOneRecognizeAuthenticateCallback`
  - https://github.com/gabrielepaolocci/ping-ios-sdk/commit/088f11c19112816a08317e81343e7c42f72ef4f4  `AUTHENTICATE` always routes to
  the Authenticate callback now (matches Android's factory; the enrollment-restore decision lives inside `authenticate()`, not at
  routing time)
  - https://github.com/gabrielepaolocci/ping-ios-sdk/commit/b81cfe157f265dc5612f3e35bca0e1320a018b77  `execute()` renamed to `enroll()`
  / `authenticate()` (also matches every other iOS callback in the SDK, Fido, Protect, ExternalIdP all use operation-specific names),
  `RecognizeError.debuggingInfo` added, `retrieveSelfie` made an explicit app-level config parameter instead of server-driven, and
  `RecognizeResult` renamed to `RecognizeSuccess`

  Let me know if this looks good to you.

Recognize/Recognize.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,605 @@
...
+/* End PBXContainerItemProxy section */
+
+/* Begin PBXCopyFilesBuildPhase section */
+		CA001015FACE000000000001 /* Embed Frameworks */ = {
+			isa = PBXCopyFilesBuildPhase;
+			buildActionMask = 2147483647;
+			dstPath = "";
+			dstSubfolderSpec = 10;
+			files = (
+				CA001004FACE000000000001 /* PingJourneyPlugin.framework in Embed Frameworks */,
```
george-bafaloukas-forgerock, line 51: This embeds `PingJourneyPlugin.framework` and `PingOrchestrate.framework` directly in `PingRecognize.framework`'s own "Embed Frameworks" phase, which is different from how our other umbrella modules do this (they typically only embed device-specific/dynamic third-party deps — e.g. `KeylessSDK` here, per the "KeylessSDK is a dynamic framework" commit). Since every consumer of this SDK already links `PingJourneyPlugin`/`PingOrchestrate` directly, embedding them again inside `PingRecognize.framework` risks duplicate-framework/duplicate-symbol errors at the app level. Can we drop these two from the Embed Frameworks phase and just link them normally?

Package.swift
```diff
@@ -315,6 +317,26 @@ let package = Package(
...
+        .target(
+            name: "PingRecognize",
+            dependencies: [
+                "PingJourneyPlugin",
+                .product(name: "KeylessSDK", package: "keyless.mobile-sdk", condition: .when(platforms: [.iOS]))
+            ],
+            path: "Recognize/Recognize",
+            exclude: ["Recognize.h"],
+            resources: [.copy("PrivacyInfo.xcprivacy")],
+            swiftSettings: [.swiftLanguageMode(.v6)]
```
george-bafaloukas-forgerock, line 331: `PingRecognize` opts into `.swiftLanguageMode(.v6)`, but the `RecognizeTests` target below doesn't set the same `swiftSettings`. That means the test target (including the `Testable*Callback` subclasses overriding `@unchecked Sendable` async methods) isn't compiled under the same strict-concurrency mode as production code — a Swift 6 concurrency violation in `PingRecognize` could pass tests while still failing for real consumers on `.v6`. Worth adding `swiftSettings: [.swiftLanguageMode(.v6)]` to the test target too so we're testing under the same mode we ship.

Recognize/Recognize/Journey/AbstractRecognizeCallback.swift
```diff
@@ -0,0 +1,504 @@
...
+    public var cameraDelaySeconds: Int? { raw[JourneyConstants.cameraDelaySeconds].flatMap(Int.init) }
+
+    /// Whether to show a success feedback overlay after the operation completes.
+    public var showSuccessFeedback: Bool? { raw[JourneyConstants.showSuccessFeedback].flatMap(Bool.init) }
+
+    // MARK: Enrollment-only options
+
+    /// A custom secret to bind to the enrollment record.
+    /// - Note: Parsed from the server but not yet forwarded to the Keyless SDK — behaviour under investigation.
+    public var customSecret: String { raw[JourneyConstants.customSecret] ?? "" }
```
george-bafaloukas-forgerock, line 63: `customSecret` is parsed off the server payload but I don't see it forwarded anywhere in `performEnroll`/`performAuthenticate` to the Keyless SDK calls — it looks like it's silently dropped today. Given this is meant to be a secret bound to the enrollment record, can we either wire it through, or explicitly log/reject when it's present but unsupported, rather than quietly ignoring it? (CodeRabbit flagged this too.)

gabrielocci: Thanks for pointing this out. The apparent omission came from a misunderstanding exposing customSecret as a  RecognizeMobileSDKOptions property  in the initial iOS implementation.

After comparing the implementation with the current Android Recognize integration, I removed customSecret from the typed iOS options instead of mapping it to savingSecret. Android keeps mobileSDKOptions generic and does not expose or forward customSecret; savingSecret is a separate Keyless secret-storage concept and is not an equivalent replacement.

The iOS implementation is now aligned with Android: unsupported/unknown server options are not exposed as part of the public typed options and are ignored, while the supported enrollment options continue to be mapped to BiomEnrollConfig.

This has been addressed here https://github.com/gabrielocci/ping-ios-sdk/commit/4ecd4d2343165a882d834093b83c63e7864d2aa7

Recognize/Recognize/RecognizeResult.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    /// The client state blob produced by the Keyless SDK, if client-state generation was enabled.
+    public let clientState: String?
+
+    /// The Recognize user ID assigned during enrollment. Empty for authentication operations.
+    public let recognizeId: String?
+
+    /// The captured selfie frame, present only when `retrieveSelfie` was `true` in `mobileSDKOptions`.
+    public let selfie: CGImage?
+}
```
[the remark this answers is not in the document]
gabrielocci: Resolved. RecognizeResult was renamed to RecognizeSuccess, which now exposes an explicit public initializer accepting signedJwt, clientState, recognizeId, and selfie. The documentation also clarifies that recognizeId is nil for authentication and explains the explicit @unchecked Sendable handling for CGImage.

SampleApps/PingExample/PingExample.xcodeproj/project.pbxproj
```diff
@@ -2383,6 +2442,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_ASSET_PATHS = "\"PingExample/Preview Content\"";
+				DEVELOPMENT_TEAM = GV4JX83LNW;
```
george-bafaloukas-forgerock: The sample project should not hard-code a specific development team or change its established bundle identifier. With automatic signing, `GV4JX83LNW` requires external contributors to belong to that team, and `com.pingi.PingExampl` breaks the existing app identity (including any configured redirect/service registrations). Please restore `com.pingidentity.PingExample` and leave the development team unset or configurable.

gabrielocci: Thanks for pointing this out. The sample project had local development signing settings and an incorrect bundle identifier carried into the PR merge by mistake. 
I removed the hardcoded development team and restored the standard com.pingidentity.PingExample bundle identifier for Debug and Release, so consumers can configure signing for their own development team without inheriting mine.

Fixed here https://github.com/gabrielocci/ping-ios-sdk/commit/f63d85aa3994fd0b24ec462521182c7e55b87b0d

Package.swift
```diff
@@ -1,13 +1,43 @@
...
+    // evaluation, so reading the user's SwiftPM configuration is unreliable;
+    // a marker file inside the package (readable in every sandbox) plus an
+    // environment variable override cover both Xcode and CLI usage.
+    let markerPath = URL(fileURLWithPath: #file).deletingLastPathComponent()
+        .appendingPathComponent("Recognize/.registry-enabled").path
+    if FileManager.default.fileExists(atPath: markerPath) {
+        return true
+    }
+    if ProcessInfo.processInfo.environment["MANIFEST_DEBUG_FATAL"] == "1" {
+        fatalError("MANIFEST-DEBUG-FATAL")
```
george-bafaloukas-forgerock: Small nit, not a blocker: this `MANIFEST_DEBUG_FATAL` check is unreachable in any normal environment (nobody sets that env var by accident), but it does live in the shared manifest and runs on *every* consumer's SPM resolution, not just when building Recognize. Since it doesn't seem to be referenced by any test, doc, or CI config, would it be worth just dropping it before merge so we don't ship debug scaffolding in `Package.swift`? 🙂

Recognize/Recognize/Journey/AbstractRecognizeCallback.swift
```diff
@@ -0,0 +1,498 @@
...
+    private(set) public var mobileSDKOptions: RecognizeMobileSDKOptions = RecognizeMobileSDKOptions(raw: [:])
+
+    // MARK: - Initialization
+
+    /// Populates output properties from the JSON output fields delivered by the server.
+    public override func initValue(name: String, value: Any) {
+        switch name {
+        case JourneyConstants.operationType:
+            if let stringValue = value as? String {
+                self.operationType = RecognizeOperationType(rawValue: stringValue) ?? .enroll
```
george-bafaloukas-forgerock: Circling back on this one — I don't think fb894f7d actually landed a fix here. That commit only flips `performAuthenticate`/`configure` from `public` to `open`; it doesn't touch this `initValue` case at all, and `stringValue` still isn't `.uppercased()` before the `RawRepresentable` lookup (unlike the other server strings in this file). So an unrecognized or mis-cased `operationType` still silently falls through to `.enroll` — and `testInitValueOperationTypeUnknownDefaultsToEnroll` even locks that in as the expected behavior. Given the enroll-vs-authenticate mixup this could cause, could we get an `.uppercased()` (or better, fail closed / surface an error) before merge? Happy to pair on it if useful!

gabrielocci: Good catch, thanks. Fixed now ([here](https://github.com/gabrielocci/ping-ios-sdk/commit/b3334956ba0e9dd00e959cdf5005c0fd309a16d6)): operationType no longer defaults to .enroll when it's missing or unrecognized. enroll() / authenticate() now check it first and fail closed with a RecognizeError before calling the Keyless SDK at all.

While fixing this, I also looked at how the Android SDK handles the same field and moved iOS closer to it: no more case normalization (exact match only, same as Android), and two separate error messages, one for a missing value, one for an unsupported value, with wording that matches Android's.

One thing I want to be upfront about: on iOS the failure still happens on the first enroll() / authenticate() call, not at parse time like on Android. That's because AbstractCallback.initialize(with:) isn't throws, it's shared by every Journey callback in the SDK, not just Recognize, so making it throw would be a bigger change across the whole JourneyPlugin module.

Package.swift
gabrielocci: Agreed, dropped it [here](https://github.com/gabrielocci/ping-ios-sdk/commit/e4d9d5612a65514b2c33a9a63d2ac9a6410b3ca1). I checked the whole repo and it wasn't referenced by any test, doc, or CI config, so it was safe to remove. I also fixed the comment above it, which described an environment-variable override that never actually worked that block only crashed on purpose for debugging, it never enabled the registry.

george-bafaloukas-forgerock (approved):
LGTM

Package.swift
```diff
@@ -1,13 +1,40 @@
...
+// Swift package registry (Cloudsmith). The dependency is only included when
+// that registry is configured on the machine, so the rest of the SDK and the
+// sample apps keep building for users without registry credentials.
+let hasKeylessRegistry: Bool = {
+    // Detects a configured Keyless registry: a marker file inside the package
+    // (checked first, since it's readable even inside xcodebuild's manifest
+    // evaluation sandbox) or, as a fallback for CLI usage, the user's SwiftPM
+    // registries.json configuration.
+    let markerPath = URL(fileURLWithPath: #file).deletingLastPathComponent()
+        .appendingPathComponent("Recognize/.registry-enabled").path
```
vahancouver: should `Recognize/.registry-enabled` file be added to `gitignore` so we don't accidentally commit it and flip `hasKeylessRegistry` to true for everyone?

Recognize/Recognize/RecognizeSuccess.swift
```diff
@@ -0,0 +1,44 @@
...
+public struct RecognizeSuccess: @unchecked Sendable {
+
+    /// The signed JWT produced by the Keyless SDK, if JWT signing was configured.
+    public let signedJwt: String?
+
+    /// The client state blob produced by the Keyless SDK, if client-state generation was enabled.
+    public let clientState: String?
+
+    /// The Recognize user ID assigned during enrollment. `nil` for authentication operations.
+    public let recognizeId: String?
```
vahancouver: doc says it's "nil for authentication operations," but performAuthenticate() populates it from Keyless.getUserId() — doc contradicts implementation.

vahancouver (approved):
Approved with a comment and a question

spetrov:
@gabrielocci, `PingWithRecognize.xcworkspace` opened as a blank window in Xcode. It looks like this is because the workspace manifest was never committed due to `.gitignore` rule. 
Also, the `Ping` workspace shows `PingExampleWithRecognize` target, which is not supposed to there...:
image

Recognize/Recognize/RecognizeSuccess.swift
gabrielocci: Thank you, good catch, fixed [here](https://github.com/gabrielocci/ping-ios-sdk/commit/92d54e1735e2e90df4c4a74353c1ec2b28c3a418).
The doc now matches the implementation: during enrollment
`recognizeId` comes from the enrollment result, while on authentication it is read from the Keyless SDK.
The doc comment was simply outdated, the implementation intentionally returns the user ID onauthentication too.

Package.swift
gabrielocci: Good point, fixed [here](https://github.com/gabrielocci/ping-ios-sdk/commit/2013b4ea4f35cbb3269c951ba4ffb4b480e476cc). 
`Recognize/.registry-enabled` is now in `.gitignore`, so the machine-local registry state can no be committed by accident and flip hasKeylessRegistry for everyone.
The marker stays as a local-only opt-in; without it (or a registries.json), SPM consumers resolve without the Cloudsmith registry.

gabrielocci:
> @gabrielocci, `PingWithRecognize.xcworkspace` opened as a blank window in Xcode. It looks like this is because the workspace manifest was never committed due to `.gitignore` rule. Also, the `Ping` workspace shows `PingExampleWithRecognize` target, which is not supposed to there...: image


Hei Stoyan, thanks for catching both issues!

1. Workspace blank on fresh clone, fixed [here](https://github.com/gabrielocci/ping-ios-sdk/commit/58a12680cea40e13a83be8026c71003b9ab96bb9).
My fault: the legacy `*.xcworkspacedata` gitignore rule silently dropped the workspace manifest, so `PingWithRecognize.xcworkspace` was committed with only its `Package.resolved` inside, and Xcode opened a blank window on a fresh clone. The manifest is now committed (same treatment as the `Ping.xcworkspace` sibling). After pulling, opening `SampleApps/PingWithRecognize.xcworkspace` and building the `PingExampleWithRecognize` scheme should work as documented.

2. The `PingExampleWithRecognize` target showing up in `Ping.xcworkspace` is a side effect of `PingExample.xcodeproj` being shared between the two workspaces: the sample code is 100% shared (only the target differs), so the Recognize target lives in the same project. Xcode shows all targets of all projects included in a workspace, and there's no per-workspace target visibility setting, the only way to remove it would be splitting the Recognize target into a separate project. 
Building that target from the core workspace fails by design (it needs `PingRecognize.framework`, which the core workspace doesn't build), but nothing else is affected. For this PR we chose the lowest-impact approach; we'll look at a proper target split as a follow-up.

